PR #2, opened 2016-07-26
Fix X-Forwarded-For modifcation in proxy requests

queimadus:
A duplicated `X-Forwarded-For` header was being added in proxy requests if that header already existed.

This PR also simplifies the directives in `ClientIPDirectives`.

### DIFF
--- a/apso/src/main/scala/eu/shiftforward/apso/spray/ProxySupport.scala
+++ b/apso/src/main/scala/eu/shiftforward/apso/spray/ProxySupport.scala
@@ -52,7 +52,7 @@ trait ProxySupport extends ClientIPDirectives {
   }
 
   private val optionalRemoteAddress: Directive1[Option[RemoteAddress]] =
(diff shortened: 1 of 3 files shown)


andrebeat:
LGTM! 👍

apso/src/main/scala/eu/shiftforward/apso/spray/ProxySupport.scala
```diff
@@ -29,23 +30,40 @@ trait ProxySupport extends ClientIPDirectives {
...
+      case `X-Forwarded-For`(ips) :: tail =>
+        `X-Forwarded-For`(ips :+ ip) :: tail
+
+      case notForwardedFor :: tail =>
+        notForwardedFor :: addForwardedFor(ip, tail)
+    }
+  }
+
+  private val optionalRemoteAddress: Directive1[Option[RemoteAddress]] =
+    headerValuePF { case `Remote-Address`(address) ⇒ Some(address) } | provide(None)
```
andrebeat: Please remove the fancy unicode arrow.

PR #3, opened 2016-08-01
Daemonize S3Bucket.TransferManager threadpool

andrebeat:
This PR changes the default thread pool executor of `S3Bucket`'s `TransferManager` to use daemon threads.

Currently the `FileDescriptor` does not provide any method to "close" the file and free any underlying resources (and also on `S3Bucket`). Using the default `TransferManager` thread pool would leave the JVM hanging waiting for those threads to finish.


queimadus:
LGTM

PR #4, opened 2016-08-01
Fix handling of nested configs in toMap method

jcazevedo:
This PR makes the implementation of the `toMap` method of `ApsoConfig` properly handle nested config objects, joining the keys with dots. This is consistent with the previous behaviour of the `toMap` method (from v0.6), while retaining the current way of handling escaped strings.


ruippeixotog:
Can you add a test (or modify the one you added) so that you include escaped strings together with nested maps?

jcazevedo:
> Can you add a test (or modify the one you added) so that you include escaped strings together with nested maps?

Sure. Just added an escaped string in a nested config.

JD557:
LGTM!

PR #7, opened 2016-08-04
Improve unit test coverage

JD557:
Updates the unit tests so that they cover more cases. This increases the test coverage to ~60%.

Also fixes some small issues that were caught by the new tests.

### DIFF
--- a/apso/src/main/scala/eu/shiftforward/apso/json/ExtraJsonProtocol.scala
+++ b/apso/src/main/scala/eu/shiftforward/apso/json/ExtraJsonProtocol.scala
@@ -3,7 +3,7 @@ package eu.shiftforward.apso.json
 import java.net.URI
 
 import scala.concurrent.duration._
-import scala.util.Try
+import scala.util.{ Failure, Success, Try }
 
 import com.github.nscala_time.time.Imports._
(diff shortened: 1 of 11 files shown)


apso/src/main/scala/eu/shiftforward/apso/json/ExtraJsonProtocol.scala
```diff
@@ -89,7 +91,11 @@ trait ExtraHttpJsonProtocol {
 trait ExtraMiscJsonProtocol {
   implicit object ConfigJsonFormat extends JsonFormat[Config] {
     def write(conf: Config): JsValue = conf.root.render(ConfigRenderOptions.concise()).parseJson
-    def read(json: JsValue): Config = ConfigFactory.parseString(json.toString)
+    def read(json: JsValue): Config = try {
+      ConfigFactory.parseString(json.toString)
+    } catch {
+      case t: Throwable => deserializationError("Could not parse config: " + json, t)
+    }
```
jcazevedo: Any reason for the explicit `try ... catch` here and using `Try` [here](https://github.com/ShiftForward/apso/pull/7/files#diff-f1cf56ee5fa9ed57f13febbb7187be73R85)?

PR #9, opened 2016-08-10
Fix/config to map

belerophon:
This changes how `toMap` handles the map keys building.

Before this change, it was possible for the method to create keys that had no correspondence in a path on the original config object. This happened because all the path segments were unquoted even if containing special characters.

A consequence of this behavior is that calling `.hasPath` on the original config object using one the produced map keys could throw an exception.

This PR uses `ConfigUtil.joinPath` to correctly quote all the path segments and join them to create the final map keys, thus guaranteeing that all the produced map keys work as paths for the original config object.

EDIT: as @jcazevedo commented, a previous implementation already existed. I've reverted back to the old implementation

### DIFF
--- a/apso/src/main/scala/eu/shiftforward/apso/config/Implicits.scala
+++ b/apso/src/main/scala/eu/shiftforward/apso/config/Implicits.scala
@@ -275,25 +275,10 @@ object Implicits extends BasicConfigReaders with ExtendedConfigReaders {
      * @tparam T the return type of the Map
      * @return the Map value
      */
-    def toMap[T](implicit configReader: ConfigReader[T]): Map[String, T] = {
-      // config key strings are different between `conf.root.entrySet` and `conf.entrySet()`
-      val unwrappedSet = conf.root.entrySet().map(e => e.getKey -> e.getValue)
-      val wrappedSet = conf.entrySet().map(e => e.getKey -> e.getValue)
-
-      unwrappedSet.flatMap {
-        case (k, v) =>
-          v match {
-            case obj: ConfigObject => obj.toConfig().toMap(configReader).map {
-              case (nk, nv) =>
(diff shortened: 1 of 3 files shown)


apso/src/main/scala/eu/shiftforward/apso/config/Implicits.scala
```diff
@@ -276,23 +276,28 @@ object Implicits extends BasicConfigReaders with ExtendedConfigReaders {
...
-      val wrappedSet = conf.entrySet().map(e => e.getKey -> e.getValue)
-
-      unwrappedSet.flatMap {
-        case (k, v) =>
-          v match {
-            case obj: ConfigObject => obj.toConfig().toMap(configReader).map {
-              case (nk, nv) =>
-                (k + "." + nk) -> nv
+
+      def aux(conf: Config, configReader: ConfigReader[T]): Map[Seq[String], T] = {
```
jcazevedo: You don't need to pass `configReader` around. It's always going to be the same under the `toMap` method.

jcazevedo:
This change reverts the changes introduced by d0f3a39c7586c9245534832f6d7ef8f7cd050472. I'm not sure why we wanted the unescaped strings in the first place, but if we want to have this behaviour now, there's a simpler implementation (check the initial `toMap` method implementation).

belerophon:
The  type safe configs requires quoted strings for certain keys. If clients of this `toMap` later try to use the created map, they too have to use quoted keys to access the values, even tough the restrictions that apply to config keys are not present on `Map` keys. So, this might be unnatural on certain situations and might explain why we've done that change.

Even so, I argue that this should be handled by the clients. `ConfigUtil.splitPath` allows one to easily unquote a complex path if required....

ruippeixotog:
LGTM!

PR #11, opened 2016-09-06
Add PureConfig support

jcazevedo:
This PR adds [PureConfig](https://github.com/melrief/pureconfig) support to `ApsoConfig`, so that it now supports the `to[T]` method for types which have the `ConfigConvert` typeclass.

We might consider deprecating the methods that use the `ConfigReader` typeclass in the future.


apso/src/main/scala/eu/shiftforward/apso/config/Implicits.scala
```diff
@@ -328,8 +330,10 @@ object Implicits extends BasicConfigReaders with ExtendedConfigReaders {
         value <- Try(conf.getConfig(key)).toOption
       } yield { key -> value }).toMap
     }
-  }
 
+    def to[T: ConfigConvert]: Try[T] =
+      loadConfig[T](pcConf.typesafeConfigToConfig(conf))
```
ruippeixotog: You can do just `loadConfig(conf)` here.

jcazevedo: Yeah. This was only recently fixed in PureConfig. I'll update it.

jcazevedo: Fixed in 53c7ff9.

ruippeixotog:
Why don't we deprecate methods using `ConfigReader` now? That would encourage people to update to PureConfig based methods gradually.

jcazevedo:
Let's leave the deprecation for later, when we've fully adopted `PureConfig`.

ruippeixotog:
Roger! LGTM then

PR #14, opened 2016-09-20
Streaming upload to file descriptors

JD557:
Adds a new `upload` method to the `FileDescriptor` that receives an `InputStream` instead of a `File`. This way, it's possible to upload data that is only stored in-memory.

Unfortunately, the S3 backend pretty much requires that the file length is known a priori, otherwise it will download the whole file to memory to find out its size. Also, the SFTP backend also seems to require the file size for logging purposes.

Therefore, this new method also receives an optional `length` attribute that it is recommended to be defined for streams with known length.

### DIFF
--- a/apso/src/main/scala/eu/shiftforward/apso/io/FileDescriptor.scala
+++ b/apso/src/main/scala/eu/shiftforward/apso/io/FileDescriptor.scala
@@ -52,6 +52,16 @@ trait FileDescriptor {
   def upload(localTarget: LocalFileDescriptor): Boolean
 
   /**
+   * Uploads an input stream to this file's location.
+   * The target must point to a file.
+   *
+   * @param inputStream the input stream that should be uploaded.
+   * @param length the input stream length (leaving this as `None` can have performance implications)
+   * @return `true` if the upload was successful, `false` otherwise.
+   */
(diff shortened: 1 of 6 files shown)


apso/src/main/scala/eu/shiftforward/apso/io/FileDescriptor.scala
```diff
@@ -52,6 +52,16 @@ trait FileDescriptor {
...
 
   /**
+   * Uploads an input stream to this file's location.
+   * The target must point to a file.
+   *
+   * @param inputStream the input stream that should be uploaded.
+   * @param length the input stream length (leaving this as `None` can have performance implications)
+   * @return `true` if the upload was successful, `false` otherwise.
+   */
+  def upload(inputStream: InputStream, length: Option[Long]): Boolean
```
jcazevedo: What do you think of adding a default value of `None` to length? I understand the performance implications of having it as `None` in some implementations, but forcing the call to `upload` to have a `length` argument can be cumbersome for implementations that don't really need it.

JD557: I would rather leave this as is for now, as not defining the length when uploading to S3 can lead to OOM errors.

We can eventually change this if we add new file descriptors that do not require the length.

PR #15, opened 2016-09-27
akka-http reverse proxy

ruippeixotog:
This pull request ports our Spray-based reverse proxy routes to akka-http.


queimadus:
LGTM

PR #18, opened 2016-10-13
Make json.Implicits.fromFullPaths separator customizable

JD557:
Changes the `json.Implicits.fromFullPaths` method to also accept an optional separator regex, instead of always using `"\\."`.

`"\\."` is still the default value, so this should be a backwards compatible change.


jcazevedo:
LGTM

PR #19, opened 2016-10-14
Increase proxy support tests timeouts

JD557:
The `ProxySupportSpec`s kept failing remotely due to timeouts.

This PR increases the timeouts of both tests and changes the specification title to make it easier to distinguish the spray proxy results from the akka-http proxy results-


jcazevedo:
LGTM

PR #21, opened 2016-10-18
InsistentInputStream performance improvements

JD557:
The `InsistentInputStream`had some performance problems with bulk reads (it wrapped every Byte read in a `Try`).

This becomes a problem when using `StreamConverters.fromInputStream` from akka-streams, which has a default chunk size of 8192 Bytes.

### DIFF
--- a/apso/src/main/scala/eu/shiftforward/apso/io/InsistentInputStream.scala
+++ b/apso/src/main/scala/eu/shiftforward/apso/io/InsistentInputStream.scala
@@ -6,6 +6,8 @@ import scala.annotation.tailrec
 import scala.concurrent.duration._
 import scala.util.{ Failure, Success, Try }
 
+import eu.shiftforward.apso.Logging
+
 /**
  * A InputStream that wraps another InputStream, retrying failed reads. This is useful for input streams that can have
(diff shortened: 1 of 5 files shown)


apso/src/main/scala/eu/shiftforward/apso/io/InsistentInputStream.scala
```diff
@@ -36,21 +38,37 @@ class InsistentInputStream(streamBuilder: () => InputStream, maxRetries: Int = 1
...
-          readRetries(nextRemainingTries)
+          readRetries(nextRemainingTries, f)
         }
     }
   }
 
-  def read(): Int = readRetries(maxRetries)
+  def read(): Int = {
+    val nextByte = readRetries(maxRetries, innerStream.read())
+    if (nextByte > 0) currPos += 1
```
andrebeat: Shouldn't this be `nextByte >= 0`?

JD557: You are correct (I messed up when I was copying code from the other implementations).

Fixed in 04e1b50.

ruippeixotog:
LGTM!

PR #24, opened 2016-11-21
Change Buffered composite iterator logic

JD557:
Some iterators (eg. iterators that use a scratch memory) might have custom logic on their buffered versions that was being ignored by our old composite iterator.

This fixes that by moving the responsibility of storing the head (if needed) to the inner iterators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,18 +6,18 @@ Apso is ShiftForward's utilities library. It provides a series of useful methods
 
 ## Installation
 
-Apso's latest release is `0.9.5` and is built against Scala 2.11.8.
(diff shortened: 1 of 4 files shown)


JD557:
The scala Iterator also stores a head on `dropWhile` and `collect` instead of calling `buffered` (to avoid an extra layer of wrapping). Do you think we should also override those methods?

ruippeixotog:
Yes, I think it's better to override those methods. It will always be susceptible to future implementation changes on `Iterator`, but at least for now it will avoid strange behavior by accidentally using some `Iterator` methods.

ruippeixotog:
LGTM

andrebeat:
👍

PR #26, opened 2016-11-25
S3Bucket: when finding files with a matching prefix use debug level logging

andrebeat:
This is a very noisy message that should be logged as debug.


ruippeixotog:
LGTM

PR #35, opened 2017-07-04
Feature/new cache directive

belerophon:
New cache directive that accepts a finite duration instead of a long.

### DIFF
--- a/apso/src/main/scala/eu/shiftforward/apso/akka/http/ExtraMiscDirectives.scala
+++ b/apso/src/main/scala/eu/shiftforward/apso/akka/http/ExtraMiscDirectives.scala
@@ -8,7 +8,7 @@ import scala.util.Try
 import akka.http.scaladsl.model.headers.CacheDirectives._
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.server.Directives._
-import akka.http.scaladsl.server.{ Directive0, Directive1 }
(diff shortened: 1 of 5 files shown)


apso/src/main/scala/eu/shiftforward/apso/akka/http/ExtraMiscDirectives.scala
```diff
@@ -17,7 +18,9 @@ trait ExtraMiscDirectives {
   final val cacheControlNoCache: Directive0 =
     respondWithDefaultHeader(`Cache-Control`(`no-cache`, `no-store`, `must-revalidate`))
 
-  def cacheControlMaxAge(inMinutes: Option[Long]): Directive0 =
+  @deprecated("Use `eu.shiftforward.apso.akka.http.ExtraMiscDirectives.cacheControlMaxAge(Option[FiniteDuration])` " +
+    "instead", "2017/07/04")
+  def cacheControlMaxAge(inMinutes: Option[Long])(implicit d: DummyImplicit): Directive0 =
```
belerophon: This does not quite work as I expected, 'cause now I have to call this as `cacheControlMaxAge(client.browserRequestCacheDuration).apply{...}`...

:(

belerophon:
@ruippeixotog I've deleted the deprecated method.

ruippeixotog:
LGTM!

PR #37, opened 2017-07-14
Deprecate old code

ruippeixotog:
This PR deprecates old code that was not being used on any ShiftForward project. Incidentally, it also changes `FileDescriptor` to use config domain objects instead of Typesafe `Config` instances, which will require some changes in the upstream projects.


core/src/test/scala/eu/shiftforward/apso/io/FileDescriptorCredentialsSpec.scala
```diff
@@ -92,22 +54,6 @@ class FileDescriptorCredentialsSpec extends Specification with CustomMatchers {
...
-      val path = "b"
-      val creds = fdCredentials.read(baseConfig, path)
-      creds must beSome(("ids-username", "ids-password"))
-
-      val path2 = "a"
-      val creds2 = fdCredentials.read(baseConfig, path2)
-      creds2 must beNone
-    }
-
-    "if the fd's is matched multiple times in the config, read the first appearance" in {
```
andrebeat: 😂

core/src/main/scala/eu/shiftforward/apso/io/FileDescriptorCredentials.scala
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * The config file must follow the following structure:
+ *
+ * {
+ *   default = $protocolCredentials
+ *    ids {
+ *      id1 = $protocolCredentials
+ *     id2 = $protocolCredentials
+ *    }
+ *
```
andrebeat: Please fix the example 😊

PR #43, opened 2017-11-10
Fix handling of S3Bucket region

andresilva:
Creating the client with `withForceGlobalBucketAccessEnabled` allows accessing any bucket without explicitly setting the region (or even using any region). The explicit bucket region is only used when creating a new bucket.


ruippeixotog:
LGTM, for what it's worth... x)

PR #44, opened 2017-11-14
Add skip parameter to the InsistentInputStream streamBuilder

JD557:
Changes the `streamBuilder` function accepted by the `InsistentInputStream` to be a `(skip: Long) => InputStream` function.

This allows us to create more efficient input streams in some cases where the inner input stream cannot skip efficiently, but can be created with an initial offset. For example, to stream data from an HTTP URL:

```scala
def stream(url: String): InputStream = {
  val builder = (skip: Long) => {
    val connection = new URI(url).toURL.openConnection().asInstanceOf[HttpURLConnection]
    if (skip > 0) connection.setRequestProperty("Range","bytes="+skip+"-")
    connection.getInputStream
  }
  new InsistentInputStream(builder)
}
```


ruippeixotog:
This LGTM! Can you just add tests for this new behavior?

PR #45, opened 2017-11-20
Add an offset parameter to the FileDescriptor#stream operation

JD557:
Adds an offset parameter to the stream operation.

This makes it easier to retry downloads without redownloading the whole file, while addressing some problematic `skip` implementations.

### DIFF
--- a/core/src/main/scala/eu/shiftforward/apso/io/LocalFileDescriptor.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/io/LocalFileDescriptor.scala
@@ -116,7 +116,11 @@ case class LocalFileDescriptor(initialPath: String) extends FileDescriptor with 
     }
   }
 
-  def stream() = new FileInputStream(file)
+  def stream(offset: Long = 0L) = {
+    val is = new FileInputStream(file)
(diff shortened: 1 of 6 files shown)


core/src/main/scala/eu/shiftforward/apso/io/LocalFileDescriptor.scala
```diff
@@ -116,7 +116,11 @@ case class LocalFileDescriptor(initialPath: String) extends FileDescriptor with
     }
   }
 
-  def stream() = new FileInputStream(file)
+  def stream(offset: Long = 0L) = {
+    val is = new FileInputStream(file)
+    if (offset > 0L) is.skip(offset)
```
ruippeixotog: Does `FileInputStream` deal with this operation correctly?

JD557: The `FileInputStream` skip is a `native` method. I assume that it will behave as expected, but it can depend on multiple factors.

According to https://stackoverflow.com/questions/3669598/does-fileinputstream-skip-do-a-seek, OpenJDK does a seek, so this should be fine.

PR #48, opened 2017-12-04
Add strict methods to the ProxySupport

JD557:
For some unknown reason, Akka-HTTP can sometimes leaks long open connections from the connection pool.

This PR adds new `Strict` methods that avoid this problem.

### DIFF
--- a/core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
@@ -1,5 +1,6 @@
 package eu.shiftforward.apso.akka.http
 
+import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ Future, Promise }
 import scala.util.{ Failure, Success, Try }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
```diff
@@ -54,47 +55,75 @@ trait ProxySupport extends ClientIPDirectives {
...
-    }
+  /**
+   * Proxies a single request to a destination URI.
+   * The response in not streamed, but converted to a strict entity with a set timeout.
+   *
+   * @param uri the target URI
+   * @param timeout maximum time to wait for the full response.
+   * @return a route that handles requests by proxying them to the given URI.
+   */
+  def proxySingleToStrict(uri: Uri, timeout: FiniteDuration): Route = proxy(Some(timeout)) {
```
ruippeixotog: Nitpick: don't you prefer `strictProxySingleTo`? The `ToStrict` at the end seems strange.

core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
```diff
@@ -54,47 +55,75 @@ trait ProxySupport extends ClientIPDirectives {
...
+  /**
+   * Proxies a single request to a destination base URI. The target URI is created by concatenating the base URI with
+   * the unmatched path.
+   * The response in not streamed, but converted to a strict entity with a set timeout.
+   *
+   * @param uri the target base URI
+   * @param timeout maximum time to wait for the full response.
+   * @return a route that handles requests by proxying them to the given URI.
+   */
+  def proxySingleToUnmatchedPathStrict(uri: Uri, timeout: FiniteDuration): Route = proxy(Some(timeout)) {
```
ruippeixotog: Same thing here.

PR #49, opened 2017-12-28
Add `Proxy.sendRequest` method

JD557:
Adds a `Proxy.sendRequest` method that allows us to send manual requests through the proxy.

This is useful to create health checks that test both the proxy and the final server.

### DIFF
--- a/core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
@@ -172,7 +172,7 @@ trait ProxySupport extends ClientIPDirectives {
      * Sends a manually crafted request to a destination URI.
      *
      * @param req the HTTP Request
-     * @param failOnDrop if the future should fail when the message is droped, or complete with a 503
+     * @param failOnDrop if the future should fail when the message is dropped, or complete with a 503
      * @return the request result.
      */
     def sendRequest(req: HttpRequest, failOnDrop: Boolean): Future[RouteResult] = {
@@ -183,7 +183,7 @@ trait ProxySupport extends ClientIPDirectives {
         case QueueClosed => Future.failed(new RuntimeException("Queue is completed before call!?"))
         case Dropped =>
           log.warn("Request queue for {}:{} is full", host, port)
-          if (failOnDrop) Future.failed(new RuntimeException("Droping request (Queue is full)"))
+          if (failOnDrop) Future.failed(new RuntimeException("Dropping request (Queue is full)"))
           else Future.successful(Complete(HttpResponse(StatusCodes.ServiceUnavailable)))
       }
     }
--- a/core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
@@ -169,6 +169,26 @@ trait ProxySupport extends ClientIPDirectives {
     private[this] lazy val queue = source.via(flow).toMat(sink)(Keep.left).run()
 
     /**
+     * Sends a manually crafted request to a destination URI.
+     *
+     * @param req the HTTP Request
+     * @param failOnDrop if the future should fail when the message is droped, or complete with a 503
+     * @return the request result.
+     */
+    def sendRequest(req: HttpRequest, failOnDrop: Boolean): Future[RouteResult] = {
+      val promise = Promise[RouteResult]()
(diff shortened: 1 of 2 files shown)


core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
```diff
@@ -168,6 +168,26 @@ trait ProxySupport extends ClientIPDirectives {
 
     private[this] lazy val queue = source.via(flow).toMat(sink)(Keep.left).run()
 
+    /**
+     * Sends a manually crafted request to a destination URI.
+     *
+     * @param req the HTTP Request
+     * @param failOnDrop if the future should fail when the message is droped, or complete with a 503
```
ruippeixotog: s/droped/dropped/

core/src/main/scala/eu/shiftforward/apso/akka/http/ProxySupport.scala
```diff
@@ -168,6 +168,26 @@ trait ProxySupport extends ClientIPDirectives {
...
+     */
+    def sendRequest(req: HttpRequest, failOnDrop: Boolean): Future[RouteResult] = {
+      val promise = Promise[RouteResult]()
+      queue.offer(req -> promise).flatMap {
+        case Enqueued => promise.future
+        case OfferFailure(ex) => Future.failed(new RuntimeException("Queue offering failed", ex))
+        case QueueClosed => Future.failed(new RuntimeException("Queue is completed before call!?"))
+        case Dropped =>
+          log.warn("Request queue for {}:{} is full", host, port)
+          if (failOnDrop) Future.failed(new RuntimeException("Droping request (Queue is full)"))
```
ruippeixotog: s/Droping/Dropping/

core/src/test/scala/eu/shiftforward/apso/akka/http/ProxySupportSpec.scala
```diff
@@ -153,6 +156,14 @@ class ProxySupportSpec(implicit ee: ExecutionEnv) extends Specification with Spe
         status == OK
         responseAs[String] must be_==("/remote-proxy")
       }
+
+      proxy.sendRequest(Get("/proxied"), failOnDrop = false).flatMap {
+        case Complete(res) => res.entity.toStrict(10.seconds).map { r =>
+          val charset = r.contentType.charsetOption.map(_.nioCharset()).getOrElse(Charset.forName("UTF-8"))
```
ruippeixotog: I wonder if for testing purposes you can't just call `r.data.utf8String` instead of doing the full, "correct" process.

core/src/test/scala/eu/shiftforward/apso/akka/http/ProxySupportSpec.scala
```diff
@@ -153,6 +156,14 @@ class ProxySupportSpec(implicit ee: ExecutionEnv) extends Specification with Spe
         status == OK
         responseAs[String] must be_==("/remote-proxy")
       }
+
+      proxy.sendRequest(Get("/proxied"), failOnDrop = false).flatMap {
```
ruippeixotog: Can you add tests here for the case where `failOnDrop = true` and for the cases where the request is effectively dropped? Creating a proxy with a `reqQueueSize` of zero should do the trick.

core/src/test/scala/eu/shiftforward/apso/akka/http/ProxySupportSpec.scala
```diff
@@ -153,6 +164,26 @@ class ProxySupportSpec(implicit ee: ExecutionEnv) extends Specification with Spe
...
+        case Complete(res) => Future.successful(res.status.intValue().toString)
+        case _ => Future.failed(new Exception("Failed to parse result"))
+      }
+
+      proxy.sendRequest(Get("/proxied"), failOnDrop = false)
+        .flatMap(parseResult) must be_==("/proxied").awaitFor(10.seconds)
+
+      val badProxy = new Proxy(interface, port, reqQueueSize = 1)
+      // Fill the bad proxy
+      (0 to 10).foreach(x => badProxy.sendRequest(Delete("/proxied-" + x), failOnDrop = false))
```
ruippeixotog: Shouldn't one be enough in this case?

JD557: Nope. I think that the internal Buffer stores some extra space, but it's hard for me to tell.

`(0 to 4)` seems to do the trick, but I would rather add some extra messages just in case.

PR #53, opened 2018-01-25
Custom JSON error handling

JD557:
Adds a `errorHandler: (JsValue, Throwable) => A` field to the  `JsonFormatBuilder.customJsonFormat`.

This allows us to recover some errors or add context to some error messages.

I also needed to add two type aliases:
```scala
type ReadFunc[+A] = C => A
type WriteFunc[-A] = A => C
```

This is helpful to drop the `C` in helper functions such as:
```scala
// Without type aliases
def myJsonFormat[C <: HList, A](builder: JsonFormatBuilder[C, _ <: HList])(read: C => A, write: A => C): RootJsonFormat[A]

// With type aliases
def myJsonFormat[A](builder: JsonFormatBuilder[_ <: HList, _ <: HList])(read: builder.ReadFunc[A], write: builder.WriteFunc[A]): RootJsonFormat[A]
```
which is much easier to call, as you usually need to define `A` manually when calling this function.

### DIFF
--- a/core/src/test/scala/eu/shiftforward/apso/json/JsonFormatBuilderSpec.scala
+++ b/core/src/test/scala/eu/shiftforward/apso/json/JsonFormatBuilderSpec.scala
@@ -129,11 +129,16 @@ class JsonFormatBuilderSpec extends Specification {
         { case a :: b :: c :: HNil => Test(a, b, c) },
         { test: Test => test.a :: test.b.tail :: test.c :: HNil },
         { (_: Test, obj: JsObject) => JsObject((obj.fields + ("c" -> 1.5.toJson)).toList: _*) },
-        (_: JsValue, _: Throwable) => throw new Exception("Caught error!"))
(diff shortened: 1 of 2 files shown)


core/src/test/scala/eu/shiftforward/apso/json/JsonFormatBuilderSpec.scala
```diff
@@ -115,5 +117,23 @@ class JsonFormatBuilderSpec extends Specification {
       Test4(Some(3), None).toJson(jw) mustEqual """{ "a": 3 }""".parseJson
       Test4(None, None).toJson(jw) mustEqual """{ }""".parseJson
     }
+
+    "allow defining custom JSON Formats with transformations and custom error handling" in {
```
ruippeixotog: Can you add a test here for when you recover from an error with a value, instead of a custom exception?

PR #54, opened 2018-02-27
Correct logging line

belerophon:


belerophon:
In 88ed472 I reverted the most unorthodox usages of adpated params

PR #55, opened 2018-03-06
Add useful circe Encoders and Decoders

jcazevedo:
This PR adds useful `Encoder`s and `Decoder`s for types we already had a `JsonFormat` for.


core/src/main/scala/eu/shiftforward/apso/json/ExtraJsonProtocol.scala
```diff
@@ -20,23 +21,25 @@ object ExtraJsonProtocol
   with ExtraMiscJsonProtocol
 
 trait ExtraTimeJsonProtocol {
+  def tryToParseDuration(duration: String): Try[FiniteDuration] =
```
ruippeixotog: Can you make this private? It's strange to import `ExtraJsonProtocol` expecting only implicit instances and get a method like this in scope.

jcazevedo: Sure. It makes sense. I made it private in 1980bd5.

PR #59, opened 2018-04-17
Add circe implicits

mbeatrizmagalhaes:
Add new methods to manipulate circe json objects. It adds the `flattenedKeySet()` as already existed for spray and adds two methods to get fields and delete a field from an object. These are helpers to substitute somethings that were done with `JsonLenses`.

### DIFF
--- a/core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
@@ -96,11 +96,15 @@ object Implicits {
      * @return flattened key set
      */
     def flattenedKeySet(separator: String = ".", ignoreNull: Boolean = true): Set[String] = {
-      val fields = json.as[JsonObject].toOption.get.toMap.toSet
-      fields.flatMap {
-        case (k, v) if v.isObject => v.flattenedKeySet(separator, ignoreNull).map(k + separator + _)
-        case (_, v) if v.isNull && ignoreNull => Set.empty[String]
-        case (k, _) => Set(k)
+      json.asObject match {
+        case None => Set.empty
+        case Some(jo) =>
+          val fields = jo.toMap.toSet
+          fields.flatMap {
+            case (k, v) if v.isObject => v.flattenedKeySet(separator, ignoreNull).map(k + separator + _)
+            case (_, v) if v.isNull && ignoreNull => Set.empty[String]
+            case (k, _) => Set(k)
+          }
       }
     }
 
--- a/core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
@@ -128,12 +128,15 @@ object Implicits {
      *
      * @param fieldPath path from the root of the json object to the field
      * @param separator character that separates each element of the path
-     * @return an option with the json without the deleted value
+     * @return the json without the deleted value
      */
-    def deleteField(fieldPath: String, separator: Char = '.'): Option[Json] =
-      getCursor(fieldPath, separator).delete.top
+    def deleteField(fieldPath: String, separator: Char = '.'): Json = {
+      require(fieldPath.nonEmpty, "The field path must have value.")
 
-    private[this] def getCursor(fieldPath: String, separator: Char): ACursor =
+      getCursor(fieldPath, separator).delete.top.get
+    }
+
+    def getCursor(fieldPath: String, separator: Char): ACursor =
       fieldPath.split(separator)
         .foldLeft(json.hcursor: ACursor) {
           case (cursor, field) =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
```diff
@@ -83,6 +85,58 @@ object Implicits {
...
+     * Returns a set of keys of this object where nested keys are separated by a separator character.
+     *
+     * Eg. {"a":1,"b":{"c":2},"d":null}.flattenedKeySet(".", ignoreNull = true) = Set("a","b.c")
+     *
+     * @param separator character separator to use
+     * @param ignoreNull if set, fields with a null value are ignored
+     * @return flattened key set
+     */
+    def flattenedKeySet(separator: String = ".", ignoreNull: Boolean = true): Set[String] = {
+      val fields = json.as[JsonObject].toOption.get.toMap.toSet
```
jcazevedo: This will throw if `json` can't be converted to `JsonObject`. I think we can avoid it by either making `json` be a `JsonObject`, which might make it more difficult to use this, or returning an empty `Set[String]` in case `json` is not an object.

core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
```diff
@@ -83,6 +85,58 @@ object Implicits {
...
+     * Eg. {"a":1,"b":{"c":2},"d":null}.deleteField("b.c") = {"a":1,"b":{},"d":null}
+     *
+     * @param fieldPath path from the root of the json object to the field
+     * @param separator character that separates each element of the path
+     * @return an option with the json without the deleted value
+     */
+    def deleteField(fieldPath: String, separator: Char = '.'): Option[Json] =
+      getCursor(fieldPath, separator).delete.top
+
+    private[this] def getCursor(fieldPath: String, separator: Char): ACursor =
```
jcazevedo: This method might be helpful on its own. What do you think of making it public?

core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
```diff
@@ -83,6 +85,58 @@ object Implicits {
...
+    /**
+     * Delete a field on a json object.
+     *
+     * Eg. {"a":1,"b":{"c":2},"d":null}.deleteField("b.c") = {"a":1,"b":{},"d":null}
+     *
+     * @param fieldPath path from the root of the json object to the field
+     * @param separator character that separates each element of the path
+     * @return an option with the json without the deleted value
+     */
+    def deleteField(fieldPath: String, separator: Char = '.'): Option[Json] =
```
jcazevedo: If `fieldPath` is an empty string, then this deletes the whole JSON and returns `None`. I'm not entirely sure if this is consistent with the method signature and documentation. Also, I'm not sure if this should return an `Option`. From the point of view of the cursor, the `top` should return an `Option`, but I don't see how deleting a field should not return a `Json`.

core/src/test/scala/eu/shiftforward/apso/json/ImplicitsSpec.scala
```diff
@@ -84,11 +86,65 @@ class ImplicitsSpec extends Specification {
...
+      import eu.shiftforward.apso.json.Implicits.ApsoJsonObject
+
+      val obj = parse("""{"a":"abc","b":{"c":2},"d":null}""").right.get
+      obj.getField[Int]("b.c") must beSome(2)
+      obj.getField[Int]("b,c", ',') must beSome(2)
+      obj.getField[String]("a") must beSome("abc")
+    }
+
+    "provide a method to delete a field from a JSON object" in {
+      import eu.shiftforward.apso.json.Implicits.ApsoJsonObject
```
jcazevedo: I don't think there's a need to hide this in the top-level import and only import it here.

core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
mbeatrizmagalhaes: I've changed it on 10df9c1f44a85ada171c60aac82d3b5fb404b961 to return empty `Set[String]` on conversion error

core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
mbeatrizmagalhaes: It's a good idea, there was no good reason to be private. Made public on f4c50aa3b990ea1faef20faaa49b4c2edf0efa3b

core/src/main/scala/eu/shiftforward/apso/json/Implicits.scala
mbeatrizmagalhaes: Made it return `Json`, and throw if the delete fails, and added a require to fieldPath be non empty on f4c50aa3b990ea1faef20faaa49b4c2edf0efa3b

PR #65, opened 2018-05-23
Fix s3 exists method

queimadus:
If a client does not want to get a "file not found" log error he should
call the `exists` method before calling the `pull` method.

### DIFF
--- a/core/src/main/scala/eu/shiftforward/apso/aws/S3Bucket.scala
+++ b/core/src/main/scala/eu/shiftforward/apso/aws/S3Bucket.scala
@@ -205,6 +205,15 @@ class S3Bucket(
   }.exists { _.exists(_.getKey.startsWith(key + "/")) }
 
   /**
+   * Checks whether the bucket exists
+   *
+   * @return true if the bucket exists, false otherwise.
+   */
+  def isBucket: Boolean = retry {
+    s3.doesBucketExistV2(bucketName)
+  }.getOrElse(false)
+
+  /**
    * Sets an access control list on a given Amazon S3 object.
    *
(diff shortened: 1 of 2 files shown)


core/src/main/scala/eu/shiftforward/apso/aws/S3Bucket.scala
```diff
@@ -204,6 +204,15 @@ class S3Bucket(
         .withPrefix(key)).getObjectSummaries.asScala
   }.exists { _.exists(_.getKey.startsWith(key + "/")) }
 
+  /**
+   * Checks whether the bucket exists
+   *
+   * @return true if the bucket exists, false otherwise.
+   */
+  def isBucket: Boolean = retry {
```
belerophon: This more of a `bucketExists` method, which seems to make more sense here.

core/src/main/scala/eu/shiftforward/apso/io/S3FileDescriptor.scala
```diff
@@ -111,10 +111,11 @@ case class S3FileDescriptor(
   }
 
   private lazy val isDirectoryRemote = bucket.isDirectory(builtPath)
+  private lazy val isBucketRoot = elements.isEmpty && bucket.isBucket
```
belerophon: This is more of a `isBucketAndExists` method (a given path `s3://some-bucket-name` is necessarily a pointer to a bucket but it may not exist, just as `file:///some-non-existing-dir` might be a pointer to a directory)
This method does not seem very useful on the class, we might just use its definition inside the `exists` method body.

What do you think?

PR #74, opened 2018-10-17
Different encoders and decoders for map

mbeatrizmagalhaes:
This adds encoder and decoder for map, for arrays of JSON objects.

~Depends on #73~


core/src/main/scala/com/velocidi/apso/json/ExtraJsonProtocol.scala
```diff
@@ -189,4 +190,29 @@ trait ExtraMiscJsonProtocol {
...
+   * @tparam V the types of the value of the map
+   * @return an instance of `Decoder` for the map
+   */
+  def mapJsonArrayDecoder[K: Decoder, V: Decoder]: Decoder[Map[K, V]] =
+    Decoder[List[MapEntry[K, V]]].map(_.flatMap(me => Some(me.key, me.value)).toMap)
+
+  private case class MapEntry[K, V](key: K, value: V)
+  private implicit def mapEntryEncoder[K: Encoder, V: Encoder]: Encoder[MapEntry[K, V]] = deriveEncoder[MapEntry[K, V]]
+  private implicit def mapEntryDecoder[K: Decoder, V: Decoder]: Decoder[MapEntry[K, V]] =
+    deriveDecoder[MapEntry[K, V]].validate(_.keys.map(k => k.exists(_ == "key") && k.exists(_ == "value")).get, "Expected a json object with 'key' and 'value' as keys")
```
jcazevedo: I think it's safer to do:

```suggestion
    deriveDecoder[MapEntry[K, V]].validate(_.keys.exists(k => k.exists(_ == "key") && k.exists(_ == "value")), "Expected a json object with 'key' and 'value' as keys")
```

core/src/main/scala/com/velocidi/apso/json/Implicits.scala
```diff
@@ -220,4 +220,21 @@ object Implicits {
         createJson(path.split(separatorRegex).toList, value).deepMerge(fromCirceFullPaths(rem, separatorRegex))
     }
   }
+
+  final implicit class ApsoJsonEncoder[A](val encoder: Encoder[A]) extends AnyVal {
+    /**
+     * Returns an encoder that removes the json fields that are null.
```
jcazevedo: Can you mention here that this won't recursively traverse all objects and only apply to the root?

PR #77, opened 2018-12-18
Add squants currency implicits

mbeatrizmagalhaes:
This adds implicits to help deal with `org.typelevel.squants` currencies. Some of it is temporary, due to an open PR on the library.


core/src/main/scala/com/velocidi/apso/json/ExtraJsonProtocol.scala
```diff
@@ -170,6 +171,15 @@ trait ExtraMiscJsonProtocol {
   implicit val localDateDecoder: Decoder[LocalDate] =
     Decoder[String].emapTry(v => Try(new LocalDate(v)))
 
+  implicit def currencyDecoder(implicit moneyContext: MoneyContext): Decoder[Currency] =
+    Decoder[String].emap { code =>
+      // TODO: this can be improved when https://github.com/typelevel/squants/pull/330 is accepted and released
+      val currencyOpt = moneyContext.currencies.find(_.code == code)
```
belerophon: We should be able to use the `asCurrency` method here.

mbeatrizmagalhaes: Done in 0097f80010a3564b96f213fb901834d9c8dce153

belerophon:
Tests seem to be failing because of `.toTry`. I think it does not exists in 2.11.12:
- https://www.scala-lang.org/api/2.11.12/index.html#scala.util.Either
- https://www.scala-lang.org/api/2.12.0/scala/util/Either.html

PR #88, opened 2019-08-14
Add elasticsearch bulkinserter

belerophon:
This adds support for an elasticsearch bulkinserter.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,8 +11,8 @@ lazy val core = project.in(file("core"))
   .settings(
     name := "apso",
     libraryDependencies ++= Seq(
-      "com.sksamuel.elastic4s"                    %% "elastic4s-core"                 % "7.1.2",
-      "com.sksamuel.elastic4s"                    %% "elastic4s-client-esjava"        % "7.1.2",
+      "com.sksamuel.elastic4s"                    %% "elastic4s-core"                 % "7.1.2"          % "provided",
+      "com.sksamuel.elastic4s"                    %% "elastic4s-client-esjava"        % "7.1.2"          % "provided",
       "com.amazonaws"                              % "aws-java-sdk-ec2"               % "1.11.553"       % "provided",
       "com.amazonaws"                              % "aws-java-sdk-s3"                % "1.11.553"       % "provided",
       "com.chuusai"                               %% "shapeless"                      % "2.3.3",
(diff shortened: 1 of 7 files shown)


belerophon:
Due to the cross compiling issue, I made this depend on #89

build.sbt
```diff
@@ -3,15 +3,16 @@ import ReleaseTransformations._
...
-crossScalaVersions in ThisBuild := Seq("2.11.12", "2.12.8")
+scalaVersion in ThisBuild := "2.12.9"
 
 lazy val core = project.in(file("core"))
   .dependsOn(testkit % "test")
   .settings(commonSettings: _*)
   .settings(
     name := "apso",
     libraryDependencies ++= Seq(
+      "com.sksamuel.elastic4s"                    %% "elastic4s-core"                 % "7.1.2",
```
JD557: Can we make this `"provided"`?

belerophon: Done in 0f75b15.

core/src/test/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserterSpec.scala
```diff
@@ -0,0 +1,102 @@
...
+import com.sksamuel.elastic4s.requests.searches._
+import io.circe.Json
+import io.circe.syntax._
+import net.ruippeixotog.akka.testkit.specs2.mutable.AkkaSpecification
+import org.specs2.concurrent.ExecutionEnv
+
+import com.velocidi.apso.elasticsearch.ElasticsearchBulkInserter.{ ElasticsearchDown, ElasticsearchUp, Insert }
+import com.velocidi.apso.elasticsearch.config.Elasticsearch
+
+class ElasticsearchBulkInserterSpec(implicit ee: ExecutionEnv) extends AkkaSpecification with ElasticsearchTestKit {
```
dimamo5: Can you add a test to check our retry logic?

I was thinking of something like this:
- Add a type mapping to a generic index
- Try inserting 2 docs, one that respects the mapping and one that doesn't.
- Ensure that only one doc is inserted.
- Remove the mapping
- Confirm the 2 docs were inserted correctly

belerophon: That is a good idea!

Done in a5ee658. This was done in a slightly different manner, using the "dynamic" property, because altering template fields is not allowed.

PR #89, opened 2019-08-16
Remove cross compile

belerophon:
This removes support for scala 2.11


belerophon:
> If we no longer publish for multiple versions, I think we can use the default release process of sbt-release, so there's no need to set releaseProcess. I think that the only thing we need to do is tell sbt-release that we want to publish signed releases:

Done it in 6e7dfac

PR #96, opened 2020-01-14
Make some minor improvements to the build definition

jcazevedo:
This PR makes some minor improvements to the build definition:

* Removes `DefaultArtifactVersions` in favor of explicit variables that are easier to reason about;
* Moves each module's `libraryDependencies` to their own `build.sbt`, to avoid clogging the main `build.sbt`;
* Avoids skipping the root project in the publishing step, otherwise the bundle with all the modules doesn't get published.


profiling/src/main/scala/com/velocidi/apso/profiling/CpuSampler.scala
```diff
@@ -4,9 +4,8 @@ import java.lang.management.ManagementFactory
 
 import scala.collection.mutable
 
-import org.slf4j.{ Logger, LoggerFactory }
-
 import com.velocidi.apso.profiling.CpuSampler._
+import org.apache.logging.log4j.scala.Logger
```
beatriz: Nitpick: could you organize these imports so that this is separated from the `com.velocidi` packages?

jcazevedo: Fixed it in 9c0d1ad.

PR #97, opened 2020-01-14
Cross-compile for Scala 2.13

jcazevedo:
This PR enables cross-compiling Apso for Scala 2.13. There are still some deprecations (the `-deprecation` flag isn't enabled by default), a lot of them related to Java conversions. In order to do this, I replaced log4j-api-scala by [scala-logging](https://github.com/lightbend/scala-logging), which I believe also serves our purposes. I also deprecated `CompositeIterator`. The issue of `Iterator.++` producing stack overflows was fixed in recent Scala versions, and we can no longer override the `++` method (which was made `final` since Scala 2.13).

Depends on #96.


.travis.yml
```diff
@@ -1,10 +1,11 @@
 language: scala
 scala:
-  - 2.12.9
+  - 2.12.10
+  - 2.13.1
 jdk:
   - openjdk8
 script:
-  - sbt -S-Xfatal-warnings ++$TRAVIS_SCALA_VERSION clean coverage test
```
beatriz: Why aren't we running fatal-warnings in the CI anymore?

jcazevedo: Even without the `-deprecation` flag, we still get a warning that there are deprecations, so we can't turn all warnings into errors. We can, however, enable fatal warnings for Scala 2.12. I did that in adfa9a2.

caching/src/main/scala/com/velocidi/apso/caching/Cache.scala
```diff
@@ -25,7 +25,7 @@ import scala.util.control.NonFatal
 /**
  * General interface implemented by all spray cache implementations.
  */
-trait Cache[V] { cache ⇒
```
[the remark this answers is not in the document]
jcazevedo: Yeah. Unicode characters became deprecated in Scala 2.13.

jcazevedo:
> LGTM, but I noticed we lost colours in CI, but couldn't understand why. Is there anything we can do about it?

This was related to the sbt update. Running with `-Dsbt.color=always -Dsbt.supershell=false` should fix that (0dcb0ae).

PR #99, opened 2020-01-20
Enable `sonatypeBundleRelease`

jcazevedo:
This PR enables the usage of `sonatypeBundleRelease` which was introduced in [version 3.4](https://github.com/xerial/sbt-sonatype/blob/master/ReleaseNotes.md#34) of sbt-sonatype and allows uploading artifacts to Sonatype as a bundle. Since Apso now has a bunch of artifacts to publish to Sonatype on each release, this speeds up things considerably.

I've tested this by doing a `release` up to the release step (replacing `sonatypeBundleRelease ` by ~~`;sonatypePrepare;sonatypeUploadBundle`~~ `;sonatypePrepare;sonatypeBundleUpload`).


build.sbt
```diff
@@ -96,12 +92,12 @@ releaseProcess := Seq[ReleaseStep](
   checkSnapshotDependencies,
   inquireVersions,
   runClean,
-  releaseStepCommandAndRemaining("+test"),
+  runTest,
```
belerophon: Does this run tests over all versions?

jcazevedo: Well spotted! sbt-release added built-in support for [cross building and cross publishing](https://github.com/sbt/sbt-release#cross-building-during-a-release) since version 0.7, but we need to set `releaseCrossBuild` to true, which I forgot. Added it in 642f4fb.

PR #104, opened 2020-03-20
Add CI workflow for github actions and remove travis.yml

thyandrecardoso:


.github/workflows/compile_test.yml
```diff
@@ -0,0 +1,43 @@
...
+      matrix:
+        scala: [2.12.10, 2.13.1]
+
+    env:
+      SCALA_VERSION: ${{ matrix.scala }}
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
```
dimamo5: For consistency with our other repos, do you mind changing this actions to `olafurpg/setup-scala@v7` ?

thyandrecardoso: Changed it in e6fd242.
Generally, if I don't have a specific requirement that could only be satisfied by using another action, I prefer to use official ones. But as we are using this elsewhere, I reckon it is OK.

PR #105, opened 2020-06-05
Control lifecycle of `ElasticClient` from within `ElasticsearchBulkInserter`

jcazevedo:
Disables supplying an external `ElasticClient` to `ElasticsearchBulkInserter` and makes sure its lifecycle (creation and closing) follows the lifecycle of the actor.

Incidentally, this also starts providing the complete `Throwables` to log entries, so that it becomes possible to have access to a stack trace.

### DIFF
--- a/elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
+++ b/elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
@@ -7,7 +7,6 @@ import scala.util.{ Failure, Success, Try }
 
 import akka.actor._
 import akka.dispatch.ControlMessage
-import akka.event.Logging
 import com.sksamuel.elastic4s.ElasticDsl._
 import com.sksamuel.elastic4s.requests.bulk.{ BulkResponse, BulkResponseItem }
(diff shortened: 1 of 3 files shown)


elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
```diff
@@ -21,26 +20,40 @@ import com.velocidi.apso.Logging
...
+    fallback
+  }
 
   private[this] val maxBufferSize = bulkInserterConfig.maxBufferSize
   private[this] val maxTryCount = bulkInserterConfig.maxTryCount
   private[this] val flushFrequency = bulkInserterConfig.flushFrequency
   private[this] val esDownCheckFrequency = bulkInserterConfig.esDownCheckFrequency
 
+  private[this] var client: ElasticClient = null
+  private[this] var esStateListeners = List.empty[ActorRef]
```
JD557, line 42: Can we remove this? I think this is never used.

jcazevedo: I think we can. I've removed the state listening logic in cd3182a.

PR #115, opened 2020-10-19
Use mdoc for README

beatriz:
This adds [mdoc](https://scalameta.org/mdoc) for the code fragments on the README. I also had to make some changes for them to compile.
The main motivation for this is to simplify the creation and maintenance of the code fragments on the README, so that we can make sure that they compile with our current code.

The changes must be made on the `docs/README.md` and then run on `sbt`: `docs/mdoc` which updates the `README.md` on the root.

I would like to have a check on CI to warn us if there were changes to the docs that weren't built and commit, but we currently have some randomness on the generated code, so it wouldn't work as it would always fail. I will look into it and improve on this in a future PR.


project/plugins.sbt
```diff
@@ -4,3 +4,4 @@ addSbtPlugin("org.scalariform"    % "sbt-scalariform"       % "1.8.2")
 addSbtPlugin("org.scoverage"      % "sbt-coveralls"         % "1.2.7")
 addSbtPlugin("org.scoverage"      % "sbt-scoverage"         % "1.6.1")
 addSbtPlugin("org.xerial.sbt"     % "sbt-sonatype"          % "3.8.1")
+addSbtPlugin("org.scalameta"      % "sbt-mdoc"              % "2.2.9" )
```
jcazevedo, line 7: Any reason why we're not using the most recent 2.2.10 version?

docs/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+You can also have the distance function curried if you are computing distances from a fixed point:
+
+```scala mdoc
+val distFromOffice = Geo.distanceFrom((41.1617609, -8.6024716))
+
+distFromOffice((41.1763745, -8.5964861))
+
+distFromOffice((38.7223032, -9.1414664))
+```
```
jcazevedo, line 128: We should wrap the `distFromOffice` declaration in a silent block to avoid having a change in the docs with an instance that doesn't have a stable `toString`.

```diff
diff --git a/docs/README.md b/docs/README.md
index 0cb0733..ec19dd2 100644
--- a/docs/README.md
+++ b/docs/README.md
@@ -119,9 +119,11 @@ Geo.distance((41.1617609, -8.6024716), (41.1763745, -8.5964861))
 
 You can also have the distance function curried if you are computing distances from a fixed point:
 
-```scala mdoc
+```scala mdoc:silent
 val distFromOffice = Geo.distanceFrom((41.1617609, -8.6024716))
+```
 
+```scala mdoc
 distFromOffice((41.1763745, -8.5964861))
 
 distFromOffice((38.7223032, -9.1414664))
```

docs/README.md
```diff
@@ -0,0 +1,613 @@
...
+val a = new A
+
+a.log.info("test")
+```
+
+### ProgressBar
+
+The `ProgressBar` represents a widget to print a dynamic progress bar in a console.
+
+```scala mdoc:reset
```
jcazevedo, line 188: This should be behind a `compile-only` block, since it depends on the time it takes to execute.

docs/README.md
```diff
@@ -0,0 +1,613 @@
...
+scala.util.Random.chooseN((0 to 15).toSeq, 4)
+
+scala.util.Random.chooseN((0 to 15).toSeq, 4)
+```
+
+### JreVersionHelper
+
+The JreVersionHelper object provides helper methods to check the two most significant parts of the JRE version at runtime:
+
+```scala mdoc:reset
```
jcazevedo, line 164: This should also probably be behind a `compile-only` block, since it depends on the platform the docs are compiled in.

docs/README.md
```diff
@@ -0,0 +1,613 @@
...
+
+(0 to 15).average
+
+Map(1 -> 2, 3 -> 6).twoWayMerge(Map(2 -> 4, 3 -> 5)) { (a, b) => b }
+
+Map(1 -> 2, 2 -> 4, 3 -> 6).twoWayMerge(Map(2 -> 2, 3 -> 5)) { (a, b) => b }
+
+Map(1 -> 2, 2 -> 3).mapKeys(_ + 1)
+
+scala.util.Random.choose((0 to 15).toSeq)
```
jcazevedo, line 147: We can easily avoid the randomness in the output by using a `Random` instance with a fixed seed instead of relying on `scala.util.Random`.

docs/README.md
```diff
@@ -0,0 +1,613 @@
...
+  config = ConfigFactory.parseString("""{
+    aws {
+      access-key = "<access-key>"
+      secret-key = "<secret-key>"
+    }
+  }"""),
+  accessKeyPath = "aws.access-key",
+  secretKeyPath = "aws.secret-key")
+
+val credentials = confProvider.getCredentials
```
jcazevedo, line 346: This section should probably be behind a `silent` block. We're not interested in showing the output and `credentials` doesn't have a stable `toString`.

docs/README.md
```diff
@@ -0,0 +1,613 @@
...
+decryptor.get.decryptToString(encryptor.get.encryptToSafeString(secretData).get)
+
+```
+
+## Hashing
+
+Apso provides utilities for various hashing functions. To use it in an existing SBT project, add the following dependency to your `build.sbt`:
+
+```scala
+libraryDependencies += "com.velocidi" %% "apso-hashing" % "@VERSION"
```
jcazevedo, line 499: Missing `@`.

```suggestion
libraryDependencies += "com.velocidi" %% "apso-hashing" % "@VERSION@"
```

build.sbt
```diff
@@ -31,6 +31,23 @@ lazy val apso = (project in file("."))
...
+lazy val docs = (project in file("apso-docs"))
+  .dependsOn(apso)
+  .settings(
+    mdocOut := baseDirectory.in(ThisBuild).value,
+
+    mdocVariables := Map(
+      "VERSION" -> "0.16.1" // This version should be set to the currently released version.
+    ),
+
+    libraryDependencies ++= Seq(Dependencies.AwsJavaSdkS3),
```
jcazevedo, line 44: Can you add a comment here stating that this is necessary because `aws-java-sdk-s3` has the `provided` scope in apso-aws?

build.sbt
```diff
@@ -31,6 +31,23 @@ lazy val apso = (project in file("."))
   .dependsOn(akkaHttp, aws, caching, collections, core, elasticsearch, encryption, hashing, io, json, profiling, time)
   .aggregate(akkaHttp, aws, caching, collections, core, elasticsearch, encryption, hashing, io, json, profiling, testkit, time)
 
+
+lazy val docs = (project in file("apso-docs"))
+  .dependsOn(apso)
+  .settings(
```
jcazevedo, line 37: Shouldn't we also use `commonSettings` for the docs project?

jcazevedo (changes requested):
Thanks for this!

The unstable output is very annoying. Even without an automated CI check, having local changes forces us to unnecessarily double-check the output. I understand if you'd prefer fixing that in a follow-up PR, but the required changes are very small. I think I pointed all of them out in the review.

project/plugins.sbt
beatriz: 2.2.10 was released, I must have started doing this before that and didn't notice 😄 Updated on c1f0597fec5714d92afa83483925f676b94127e3.

build.sbt
beatriz: Yes, fixed on 510296efd350813041469c3d5306f1392635b906.

build.sbt
beatriz: Done on b01c0f09cf6ba569e3284019e9e1f4ba3cde6070.

docs/README.md
beatriz: On db85c699bd224347d2c3365383b4a0c6f6014dfd, I've made the suggested changes (and some more) to avoid the unstable output.

beatriz:
> The unstable output is very annoying. Even without an automated CI check, having local changes forces us to unnecessarily double-check the output. I understand if you'd prefer fixing that in a follow-up PR, but the required changes are very small. I think I pointed all of them out in the review.

I hadn't thought of the annoyance this would be to whoever changed the README. Made that changes so it would be stable on db85c699bd224347d2c3365383b4a0c6f6014dfd.

README.md
```diff
@@ -547,46 +589,42 @@ creation of the underlying Cyphers.
...
+decryptor.get.decryptToString(encryptor.get.encryptToSafeString(secretData).get)
 ```
 
 ## Hashing
 
 Apso provides utilities for various hashing functions. To use it in an existing SBT project, add the following dependency to your `build.sbt`:
 
 ```scala
-libraryDependencies += "com.velocidi" %% "apso-hashing" % "0.16.0"
+libraryDependencies += "com.velocidi" %% "apso-hashing" % "@VERSION"
```
jcazevedo, line 617: Can you run `docs/mdoc` so that this gets updated?

docs/README.md
```diff
@@ -0,0 +1,627 @@
...
+
+def m() = {
+  attempts += 1
+  if (attempts > 5)
+    attempts
+  else
+    throw new Exception()
+}
+
+println(Retry.retry(10)(m))
```
jcazevedo, line 275: I suppose we can drop this `println` since the result will be printed regardless.

docs/README.md
```diff
@@ -0,0 +1,627 @@
...
+    val value = a.getAndAdd(1)
+    if (value > 5)
+      value
+    else {
+      throw new Exception()
+    }
+  }
+}
+
+Await.result(Retry.retryFuture(10)(f), 1.second)
```
jcazevedo, line 263: Since we're sure that this completes, we can be more lenient and await indefinitely.

```suggestion
Await.result(Retry.retryFuture(10)(f), Duration.Inf)
```

project/plugins.sbt
```diff
@@ -4,3 +4,4 @@ addSbtPlugin("org.scalariform"    % "sbt-scalariform"       % "1.8.2")
 addSbtPlugin("org.scoverage"      % "sbt-coveralls"         % "1.2.7")
 addSbtPlugin("org.scoverage"      % "sbt-scoverage"         % "1.6.1")
 addSbtPlugin("org.xerial.sbt"     % "sbt-sonatype"          % "3.8.1")
+addSbtPlugin("org.scalameta"      % "sbt-mdoc"              % "2.2.10" )
```
jcazevedo, line 7: Nitpick.

```suggestion
addSbtPlugin("org.scalameta"      % "sbt-mdoc"              % "2.2.10")
```

jcazevedo (changes requested):
Thanks for the changes!

I commented on some minor nitpicks below.

README.md
beatriz: Fixed on 4bf9493bac5ae546c0a4c79f0ee22bc9bde0a885.

docs/README.md
beatriz: Done on 4bf9493bac5ae546c0a4c79f0ee22bc9bde0a885.

jcazevedo (approved):
Thanks!

PR #116, opened 2020-10-22
Introduce extension method emapPrettyTry to circe's Decoder

queimadus:
Using `emapTry` to create a Decoder will result in the stack trace being serialized along with the throwable's message. 

This new method behaviour is similar to `emapTry` but in case of failure it exclusively uses the Throwable's message to create the DecodingFailure message whereas `emapTry` would use the message+stacktrace.
Related issue: https://github.com/circe/circe/issues/306

It also changes all occurrences of `emapTry` to `emapPrettyTry` in the codebase.


jcazevedo (approved):
LGTM

PR #119, opened 2020-10-31
Document JSON helpers

beatriz:
This documents the `JSON` implicits and converters. We previously had these documented for `spray-json` and just removed them, as the README was needing some improvements. Now that we have `mdoc`, this is easier to add and improve.

I also thought that, now that we don't have the `spray-json` methods, we could simplify the naming of some methods and remove the `Circe` specification on them. WDYT?

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -574,6 +574,64 @@ Apso includes a bunch of utilities to work with JSON serialization and deseriali
 libraryDependencies += "com.velocidi" %% "apso-json" % "@VERSION@"
 ```
 
+### ExtraJsonProtocol
+
+The `ExtraJsonProtocol` object combines three traits that provide extra `Encoders` and `Decoders` (of [circe](https://circe.github.io/circe/)) for some relevant types. The `Encoders` and `Decoders` are provided on each trait for the following types:
+
+* ExtraTimeJsonProtocol: `FiniteDuration`, `Interval` and `Period`;	
+* ExtraHttpJsonProtocol: `URI`;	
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -574,6 +574,64 @@ Apso includes a bunch of utilities to work with JSON serialization and deseriali
...
+"a".asJson		
+"2".asJson		
+val js1 = Json.obj(
+  "a" := 2,
+  "b" := 3,
+  "d" := Json.obj("f" := 6))
+	
+val js2 = Json.obj(
+            "c" := 4,
+            "d" := Json.obj("e" := 5))
```
jcazevedo, line 602: The output of this snippet is circe-specific. It's probably preferable to silence it in the docs, since we're mostly interested in the output of the `deepMerge` method.

docs/README.md
```diff
@@ -574,6 +574,64 @@ Apso includes a bunch of utilities to work with JSON serialization and deseriali
...
+js1.getField[Int]("d.f")
+js1.getField[Int]("x")
+
+js1.deleteField("a")
+js1.deleteField("d.f")
+js1.deleteField("x")	
+```
+
+### JsonConvert
+The `JsonConvert` object contains helpers for converting between JSON values and other structures. See the following for sample usage:
```
jcazevedo, line 623: This isn't very accurate, but since it also requires an update to `JsonConvert`'s Scaladoc, I think we can fix it in a different PR.

jcazevedo (commented):
> I also thought that, now that we don't have the `spray-json` methods, we could simplify the naming of some methods and remove the `Circe` specification on them. WDYT?

Yes. I think it'd simplify stuff if we'd remove mentions to Circe in the method names. I'd also consider renaming the package to apso-circe instead of apso-json, since everything on it is Circe-specific.

docs/README.md
beatriz: Sorry for the slow response to this, this PR got forgotten 😞  Silenced this circe-specific output on a7dbab31505f75c567c1b3bc0d4e115d26a38ac4.

PR #121, opened 2020-11-09
Update Circe to 0.13.0

jcazevedo:
This PR updates circe-core, circe-generic and circe-parser to 0.13.0. circe-parser 0.13.0 is not binary compatible with 0.12.x, so this update is important for downstream users.


duartepinto (approved):
LGTM! Thanks for this

PR #122, opened 2020-12-21
Add an `isLocal` flag to `FileDescriptor`

jcazevedo:
This adds an `isLocal` flag to `FileDescriptor` enabling quickly identifying an FD as either using to access a local file or not.

### DIFF
--- a/io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
+++ b/io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
@@ -22,7 +22,7 @@ case class LocalFileDescriptor(initialPath: String) extends FileDescriptor with 
 
   lazy val name: String = file.getName
 
-  def isLocal: Boolean = true
+  val isLocal: Boolean = true
 
   lazy val file = Paths.get(initialPath).normalize().toAbsolutePath.toFile
 
--- a/io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
+++ b/io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
@@ -22,6 +22,8 @@ case class LocalFileDescriptor(initialPath: String) extends FileDescriptor with 
 
   lazy val name: String = file.getName
 
+  def isLocal: Boolean = true
+
   lazy val file = Paths.get(initialPath).normalize().toAbsolutePath.toFile
 
   def isDirectory: Boolean = file.isDirectory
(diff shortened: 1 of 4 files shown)


jcazevedo:
The CI build was broken, so this now depends on #123.

io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
```diff
@@ -22,6 +22,8 @@ case class LocalFileDescriptor(initialPath: String) extends FileDescriptor with
 
   lazy val name: String = file.getName
 
+  def isLocal: Boolean = true
```
thyandrecardoso, line 25: Nitpick:
```suggestion
  val isLocal: Boolean = true
```

io/src/main/scala/com/velocidi/apso/io/RemoteFileDescriptor.scala
```diff
@@ -8,6 +8,7 @@ trait RemoteFileDescriptor { this: FileDescriptor =>
 
   lazy val name: String = elements.lastOption.getOrElse("")
   lazy val path: String = root + "/" + elements.mkString("/")
+  def isLocal: Boolean = false
```
thyandrecardoso, line 11: Nitpick:
```suggestion
  val isLocal: Boolean = false
```

thyandrecardoso (changes requested):

io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
jcazevedo: Is the constant extra space for every file descriptor worth it?

thyandrecardoso: Well... I think we're talking about an extra constant byte (more or less), and I am assuming there is a (very small) performance improvement by it being a `val`, but that assumption might be thrown away by some optimization on the vm... so, I am perfectly ok on not having this as a `def`.

I tend to prefer a `val` because it more accuretly depicts that this should be constant for the file descriptor in question.

thyandrecardoso (approved):

io/src/main/scala/com/velocidi/apso/io/LocalFileDescriptor.scala
jcazevedo: I suppose it's OK to use `val` for this.

thyandrecardoso (approved):

PR #125, opened 2021-01-05
Use full connection details to index into SFTP FileDescription's connectionPools

queimadus:
Just using the host to index into the `connectionPools` can cause problems if connecting to different sftp's on the same host or even if trying to access the same server with different credentials.


jcazevedo (approved):
Nice catch!

PR #127, opened 2021-02-08
Make indexing failures be sent only when retries are exhausted

thyandrecardoso:
Currently, `ElasticsearchBulkInserter` replies with a `Status.Failure` at each indexing retry, for the same message.

This changes that behavior so that `Status.Failure` is only sent when the retries are exhausted.

### DIFF
--- a/elasticsearch/src/test/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserterSpec.scala
+++ b/elasticsearch/src/test/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserterSpec.scala
@@ -116,60 +116,62 @@ class ElasticsearchBulkInserterSpec(implicit ee: ExecutionEnv) extends AkkaSpeci
     }
 
     "correctly send error back to client only when retries are exhausted" in {
+      "when the buffer maximum size is reached" in {
+        val msgIndex = "test-index-5"
+        // maxTryCount is 3
+        val bulkInserter = testBulkInserter(maxBufferSize = 2, flushFreq = 1.day)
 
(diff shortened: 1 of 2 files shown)


elasticsearch/src/test/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserterSpec.scala
```diff
@@ -85,22 +98,78 @@ class ElasticsearchBulkInserterSpec(implicit ee: ExecutionEnv) extends AkkaSpeci
...
-        be_==(5).awaitFor(5.seconds).eventually(30, 2.seconds)
+      numberOfHits(msgIndex) must be_==(5).retry(eventuallyRetries = 30)
       esClient.execute(search(msgIndex) query matchQuery("other", "dynamic_field_value")).map(_.result.totalHits) must
-        be_==(1).awaitFor(5.seconds).eventually(30, 2.seconds)
+        be_==(1).retry(eventuallyRetries = 30)
+    }
+
+    "correctly send error back to client only when retries are exhausted" in {
+
+      // NOTICE: trying on a setting in which the retries are triggered by the buffer reaching its max capacity
```
queimadus, line 120: Maybe we could create separate tests for these isolated flows instead of separating them by a "NOTICE" comment?

queimadus (approved):
LGTM

elasticsearch/src/test/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserterSpec.scala
thyandrecardoso: Sure, done in 229ff3c

PR #128, opened 2021-02-12
Update elasticsearch to version 7.10

dimamo5:
Upgrades elastic4s and `elasticsearch-cluster-runner` to version 7.10.

There are no breaking changes on features used by apso.

### DIFF
--- a/testkit/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchTestKit.scala
+++ b/testkit/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchTestKit.scala
@@ -40,7 +40,6 @@ trait NoSpecElasticsearchTestKit {
       // another es-cluster-runner instance could already be using that same port
       // By adding this Random we avoid binding to ports already in use
       .baseHttpPort(9200 + Random.nextInt(50))
-      .baseTransportPort(9300 + Random.nextInt(50))
       .numOfNode(1)
       .disableESLogger()
       .useLogger())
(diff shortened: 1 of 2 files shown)


testkit/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchTestKit.scala
```diff
@@ -40,7 +40,6 @@ trait NoSpecElasticsearchTestKit {
       // another es-cluster-runner instance could already be using that same port
       // By adding this Random we avoid binding to ports already in use
       .baseHttpPort(9200 + Random.nextInt(50))
-      .baseTransportPort(9300 + Random.nextInt(50))
```
thyandrecardoso, line 43: Do you know the reason for this being removed?

thyandrecardoso (approved):

testkit/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchTestKit.scala
dimamo5: Not sure. I couldn't find the reason for this change. 

Commit that remove this setting: https://github.com/codelibs/elasticsearch-cluster-runner/commit/51d2774618e8fbc89ef41d750c25ff32bd43a3cf

JD557: I believe that the transport port was deprecated a long time ago, and replaced with the HTTP port.

thyandrecardoso (approved):

PR #129, opened 2021-02-13
Stop using success count to determine if ES is down

thyandrecardoso:
Success count was being used to determine if the server was down. However, I think we cannot infer anything from that value, given that we don't know the original number of requests.  So, I think it is best to only rely on the Failure or Success of the bulk insert Future.

### DIFF
--- a/elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
+++ b/elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
@@ -85,20 +85,19 @@ class ElasticsearchBulkInserter(
     }
   }
 
-  private[this] def countSuccessfulMsgs(sentBuffer: Iterable[Message], bulkResponse: BulkResponse) = {
+  // returns a list of the failed requests
+  private[this] def notifyOfSuccessfulAndGetFailed(sentBuffer: Iterable[Message], bulkResponse: BulkResponse) = {
     val reqsAndRes = sentBuffer.iterator.zip(bulkResponse.items.iterator)
 
-    val (successCount, failedReqs) = reqsAndRes.foldLeft((0, List.empty[Message])) {
-      case ((count, failedReqs), (req, res)) if res.error.isDefined =>
-        (count, failedReqs ::: addRetry(req, res))
+    reqsAndRes.foldLeft(List.empty[Message]) {
+      case (failedReqs, (req, res)) if res.error.isDefined =>
+        failedReqs ::: addRetry(req, res)
 
-      case ((count, failedReqs), (req, res)) => // if res.error.isEmpty...
+      case (failedReqs, (req, res)) => // if res.error.isEmpty...
         req.sender ! res.id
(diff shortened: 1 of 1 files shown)


elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
```diff
@@ -85,20 +85,19 @@ class ElasticsearchBulkInserter(
     }
   }
 
-  private[this] def countSuccessfulMsgs(sentBuffer: Iterable[Message], bulkResponse: BulkResponse) = {
+  // returns a list of the failed requests
```
JD557, line 88: Can you use proper scaladoc here (and maybe add a type annotation to the method)?

elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
```diff
@@ -85,20 +85,19 @@ class ElasticsearchBulkInserter(
...
+  private[this] def notifyOfSuccessfulAndGetFailed(sentBuffer: Iterable[Message], bulkResponse: BulkResponse) = {
     val reqsAndRes = sentBuffer.iterator.zip(bulkResponse.items.iterator)
 
-    val (successCount, failedReqs) = reqsAndRes.foldLeft((0, List.empty[Message])) {
-      case ((count, failedReqs), (req, res)) if res.error.isDefined =>
-        (count, failedReqs ::: addRetry(req, res))
+    reqsAndRes.foldLeft(List.empty[Message]) {
```
JD557, line 92: I think this can now be replaced by a `flatMap`, which should be more efficient (since iternally the list concatenation will use a mutable builder). I think we can also get rid of the guard clause, which should avoid calling `unapply` twice on the happy path.

I haven't tested it, but I think something like this should work:

```scala
reqsAndRes.flatMap { case (req, res) =>
  if (res.error.isDefined) {
    addRetry(req, res)
  }
  else {
    req.sender ! res.id 
    tryCountMap.remove(req) 
    Nil
  }
}
```

JD557 (changes requested):
I think this makes sense. It would probably be better if we had some notion of retriable/non-retriable errors, but this seems fine for now.

Just left some small nitpicks.

elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
thyandrecardoso: Yes, that seems to make sense. Thank you! Improved in e546c98

elasticsearch/src/main/scala/com/velocidi/apso/elasticsearch/ElasticsearchBulkInserter.scala
thyandrecardoso: Yes, done in e546c98

JD557 (approved):

PR #131, opened 2021-03-02
Cleanup dependencies

jcazevedo:
Removes unused dependencies, explicitly adds dependencies required for compilation which were transitively pulled, and cleans up the scopes.


jcazevedo:
@thyandrecardoso: As discussed offline, I've moved the dependencies of akka-actor and akka-stream back to the provided scope in apso-akka-http since the requirements for an `ActorSystem` and a `Materializer` are explicit.

akka-http/build.sbt
```diff
@@ -2,7 +2,6 @@ import Dependencies._
 
 libraryDependencies ++= Seq(
```
thyandrecardoso, line 3: I think we are missing two explicit dependencies here: `com.typesafe.config` and `com.typesafe.akka.akka-http-core`.

thyandrecardoso (changes requested):

akka-http/build.sbt
jcazevedo: Yes. I think that was due to a merge gone wrong.

thyandrecardoso (approved):

PR #137, opened 2021-04-28
Update akka-http to 10.2.4

jcazevedo:


queimadus (approved):
LGTM, thanks for this!

Just to make this even more evident, to keep the exact same proxy behaviour as previous apso versions, some changes are now required by users of this library:

>In order for the client IP to be correctly propagated in `X-Forward-For` headers, the `ProxySupport` trait requires the `akka.http.server.remote-address-attribute` setting to be `on`.

This setting is `off` by default.

PR #138, opened 2021-05-05
Move time related codecs to time trait and add companion objects

thyandrecardoso:
Moves DateTime and LocalDate codecs from `ExtraMiscJsonProtocol` to `ExtraTimeJsonProtocol`, and adds companion objects for all traits.


queimadus (approved):
LGTM

PR #142, opened 2021-05-14
Update scala, dependencies and plugins

beatriz:
This PR updates the scala versions, dependencies, and plugins versions. 
I just didn't update `caffeine` and `guava` because `scalacache` still doesn't support the most recent version and was getting compilation errors.


jcazevedo:
What's the status on this? Why is it on draft?

beatriz:
> What's the status on this? Why is it on draft?

I was testing updating this on a project which depended on this and run into some problems and didn't have the availability to look into it right away. I have that availability now, will look into this to open make it ready for review.

beatriz:
I've updated the projects we have and use it with this version and everything went well, so this is ready for review.

queimadus (approved):
LGTM

PR #149, opened 2021-10-29
Improve cache in CI

jcazevedo:
We don't need to have these many separated caching steps. This lists some folders unused by the OSs we currently run builds on, but I suppose it's OK to leave them there for reference (and in case we start building somewhere else).

Incidentally, this also reformats the YAML a bit.


beatriz (approved):
LGTM! The additional yaml reformat also makes it a bit easier to read.

PR #151, opened 2021-10-31
Update `scalafmt` to 3.0.8 and apply style changes

beatriz:
Updates `scalafmt` to 3.0.6 and applies the style changes. The dialect is required, otherwise, it prints warnings, so I chose `scala213` for Scala files and `sbt1` for SBT files.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,8 +1,10 @@
-version = 2.7.5
+version = 3.0.6
 
+runner.dialect = scala213
 maxColumn = 120
 fileOverride {
   "glob:**/*.sbt" {
+    runner.dialect = sbt1
     align.preset = most
   }
   "glob:**/project/*.scala" {
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 3.0.6
+version = 3.0.8
 
 runner.dialect = scala213
 maxColumn = 120
--- a/caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
+++ b/caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
@@ -26,7 +26,6 @@ import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap
 
 object LruCache {
 
-  // # source-quote-LruCache-apply
   /** Creates a new [[ExpiringLruCache]] or [[SimpleLruCache]] instance depending on whether a non-zero and finite
     * timeToLive and/or timeToIdle is set or not.
     */
@@ -36,7 +35,6 @@ object LruCache {
       timeToLive: Duration = Duration.Inf,
       timeToIdle: Duration = Duration.Inf
   ): Cache[V] = {
-    // #
     def check(dur: Duration, name: String) =
       require(
         dur != Duration.Zero,
--- a/caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
+++ b/caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
@@ -26,7 +26,7 @@ import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap
 
 object LruCache {
 
-  //# source-quote-LruCache-apply
+  // # source-quote-LruCache-apply
   /** Creates a new [[ExpiringLruCache]] or [[SimpleLruCache]] instance depending on whether a non-zero and finite
     * timeToLive and/or timeToIdle is set or not.
     */
@@ -36,7 +36,7 @@ object LruCache {
       timeToLive: Duration = Duration.Inf,
       timeToIdle: Duration = Duration.Inf
   ): Cache[V] = {
-    //#
+    // #
     def check(dur: Duration, name: String) =
       require(
         dur != Duration.Zero,
--- a/caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
+++ b/caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
@@ -27,9 +27,8 @@ import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap
 object LruCache {
 
   //# source-quote-LruCache-apply
-  /** Creates a new [[ExpiringLruCache]] or
-    * [[SimpleLruCache]] instance depending on whether
-    * a non-zero and finite timeToLive and/or timeToIdle is set or not.
+  /** Creates a new [[ExpiringLruCache]] or [[SimpleLruCache]] instance depending on whether a non-zero and finite
+    * timeToLive and/or timeToIdle is set or not.
     */
   def apply[V](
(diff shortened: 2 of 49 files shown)


.scalafmt.conf
```diff
@@ -1,8 +1,10 @@
-version = 2.7.5
+version = 3.0.6
```
jcazevedo, line 1: Can we use 3.0.8?

```suggestion
version = 3.0.8
```

io/src/main/scala/com/velocidi/apso/io/FileDescriptor.scala
```diff
@@ -27,94 +29,111 @@ trait FileDescriptor {
...
-  /** Changes the path of the file descriptor using unix's cd syntax related to the current directory.
-    * Ex: Go back to directories: ../..
-    *     Go to child directories: foo/bar
-    *     Go to sibling directory: ../foo
-    * Absolute path changes are not supported.
-    * @param pathString the cd command
-    * @return the new file descriptor with the updated path
+  /** Changes the path of the file descriptor using unix's cd syntax related to the current directory. Ex: Go back to
+    * directories: ../.. Go to child directories: foo/bar Go to sibling directory: ../foo Absolute path changes are not
+    * supported.
```
jcazevedo, line 132: Can we replace these examples with a list to avoid reformat?

```suggestion
  /** Changes the path of the file descriptor using unix's cd syntax related to the current directory.
    *
    * Ex:
    *   - Go back to directories: ../..
    *   - Go to child directories: foo/bar
    *   - Go to sibling directory: ../foo
    *
    * Absolute path changes are not supported.
```

io/src/main/scala/com/velocidi/apso/io/SftpFileDescriptor.scala
```diff
@@ -18,33 +18,22 @@ import net.schmizz.sshj.xfer.InMemorySourceFile
...
+  * Both the username and port are optional. Additionally, the credentials config expects an object with the following
+  * format:
   *
-  * `sftp {
-  *    default = {
-  *      username = <username>
-  *      password = <password>
-  *    }
-  *  }`
+  * `sftp { default = { username = <username> password = <password> } }`
```
jcazevedo, line 30: Can we replace this by source code blocks to avoid reformat?

```suggestion
  * {{{
  * sftp {
  *   default = {
  *     username = <username>
  *     password = <password>
  *   }
  * }
  * }}}
```

io/src/main/scala/com/velocidi/apso/io/SftpFileDescriptor.scala
```diff
@@ -18,33 +18,22 @@ import net.schmizz.sshj.xfer.InMemorySourceFile
...
-  * Or if using public key authentication:
-  * `sftp {
-  *    default = {
-  *      username = <username>
-  *      keypair-file = <key filename>
-  *      passphrase = <passphrase>
-  *    }
-  *  }`
+  * Or if using public key authentication: `sftp { default = { username = <username> keypair-file = <key filename>
+  * passphrase = <passphrase> } }`
```
jcazevedo, line 33: Same as above.

```suggestion
  * Or if using public key authentication:
  *
  * {{{
  * sftp {
  *   default = {
  *     username = <username>
  *     keypair-file = <key filename>
  *     passphrase = <passphrase>
  *   }
  * }
  * }}}
```

core/src/main/scala/com/velocidi/apso/Implicits.scala
```diff
@@ -89,19 +98,20 @@ object Implicits {
...
-      *   <li>Keys present in one of thw two maps are present in the merged map;
-      *   <li>Keys in both maps are present in the merged map with a value given
-      *       by `f(thisValue, thatValue)`;
-      * </ul>
-      * @param that the map to be merged with this map
-      * @param f the function used to merge two values with the same key
-      * @return the merged map.
+    /** Merges a given map with this map. The map is constructed as follows: <ul> <li>Keys present in one of thw two
+      * maps are present in the merged map; <li>Keys in both maps are present in the merged map with a value given by
+      * `f(thisValue, thatValue)`; </ul>
```
jcazevedo, line 108: Can we avoid HTML so that this doesn't get reformatted?

```suggestion
    /** Merges a given map with this map. The map is constructed as follows:
      *   - Keys present in one of thw two maps are present in the merged map;
      *   - Keys in both maps are present in the merged map with a value given by `f(thisValue, thatValue)`;
      *
```

jcazevedo (changes requested):
Thanks for this!

.scalafmt.conf
beatriz: We can, my mistake.

caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
```diff
@@ -26,18 +26,17 @@ import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap
 
 object LruCache {
 
-  //# source-quote-LruCache-apply
-  /** Creates a new [[ExpiringLruCache]] or
-    * [[SimpleLruCache]] instance depending on whether
-    * a non-zero and finite timeToLive and/or timeToIdle is set or not.
+  // # source-quote-LruCache-apply
```
jcazevedo, line 29: This can be removed. It was only useful for Spray docs.

```suggestion
```

caching/src/main/scala/com/velocidi/apso/caching/LruCache.scala
```diff
@@ -26,18 +26,17 @@ import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap
...
+    * timeToLive and/or timeToIdle is set or not.
     */
   def apply[V](
       maxCapacity: Int = 500,
       initialCapacity: Int = 16,
       timeToLive: Duration = Duration.Inf,
       timeToIdle: Duration = Duration.Inf
   ): Cache[V] = {
-    //#
+    // #
```
jcazevedo, line 39:
```suggestion
```

core/src/main/scala/com/velocidi/apso/Implicits.scala
```diff
@@ -152,15 +170,19 @@ object Implicits {
...
-      *         otherwise. Not choosing any element can happen if the weights of the elements do not
-      *         sum up to the maximum value of `r`.
+      * @param seq
+      *   the elements to choose from
+      * @param valueFunc
+      *   the function that maps elements to weights
+      * @param r
+      *   the random value used to select the elements. If the default random value is used, the weighted selection uses
+      * 1.0 as the sum of all weights. To use another scale of weights, a random value between 0.0 and the maximum
+      * weight should be passed.
```
jcazevedo, line 180: This alignment isn't right, but it looks like a Scalafmt bug.

jcazevedo (changes requested):

jcazevedo (approved):
Thanks!

PR #153, opened 2021-12-10
Remove Log4j dependency from apso-log

jcazevedo:
Downstream dependencies can decide which logging backend to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ JreVersionHelper.jreVersion
 
 ### Logging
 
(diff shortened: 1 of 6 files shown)


beatriz (changes requested):
Both the `Logging` trait scaladoc and the README refer to the `Logger` object as `Log4j2` specific, with this change we probably want to remove that reference, no?

jcazevedo:
Yes. Nice catch.

beatriz (approved):

PR #156, opened 2022-01-12
Remove `sbt-scoverage` and `sbt-coveralls` plugins

beatriz:
We aren't using the `sbt-coveralls` to report the coverage, so I removed it.


jcazevedo (approved):

jcazevedo:
Do we need to keep sbt-scoverage then?

beatriz:
> Do we need to keep sbt-scoverage then?

I left it because it allows us to generate coverage reports locally, it could be useful. But I don't have a strong opinion on this 😄

jcazevedo:
> I left it because it allows us to generate coverage reports locally, it could be useful.

We can always set it up as a global plugin if we want to generate those reports locally. I think it's preferable to remove it.

beatriz:
I've also removed `sbt-scoverage` and updated the PR title accordingly.

PR #167, opened 2022-01-13
Add .scala-steward.conf

beatriz:
This adds the `.scala-steward.conf` file to pin the version of dependencies we don't want updated automatically, so that we can activate it. All available settings are described [here](https://github.com/scala-steward-org/scala-steward/blob/master/docs/repo-specific-configuration.md).


jcazevedo:
Shouldn't we be using the `updates.ignore` setting with no explicit version instead? Having pinned versions forces us to always update this every time we update one of these dependencies.

beatriz:
> Shouldn't we be using the `updates.ignore` setting with no explicit version instead? Having pinned versions forces us to always update this every time we update one of these dependencies.

Yes, this is true for `caffeine` and `guava`. I've left the ES dependencies pinned to the minor, as we are ok with updating the patch versions.

jcazevedo (approved):

PR #187, opened 2022-01-28
Remove TryWith in favor of scala 2.13 native resource management tools

thyandrecardoso:
Scala 2.13 offers `Using` to achieve better functionality, and it is also offered by https://github.com/scala/scala-collection-compat.


jcazevedo (approved):
Thanks!

PR #274, opened 2022-06-02
Add a Scala Steward action

jcazevedo:
Adds a Scala Steward action now that the public instance is no longer up.


dimamo5 (approved):
Thanks!

PR #313, opened 2022-07-12
Remove the Scala Steward GitHub action

jcazevedo:
VirtusLab [has revived](https://contributors.scala-lang.org/t/can-scala-maintain-a-public-scala-steward-instance/5754/10) the public Scala Steward instance, so we no longer need to have our own action.


dimamo5 (approved):
Thank you!

PR #345, opened 2022-09-14
Add option to log all warnings as errors in `ElasticsearchBulkinserter`

dimamo5:
This gives us the ability to log messages like `Elasticsearch is down...` as an error instead of a warning


joaocsf (approved):
LGTM

PR #381, opened 2022-11-16
Update actions used in CI pipeline

jcazevedo:
This updates the actions used in the CI pipeline since we have been getting some deprecation warnings (see https://github.com/adzerk/apso/actions/runs/3481258143 for an example). 

The move away from [olafurpg/setup-scala](https://github.com/olafurpg/setup-scala) is because (1) the project is likely to be archived (https://github.com/olafurpg/setup-scala/issues/49) and (2) sbt is now bundled with GitHub's runners, so we can rely on [actions/setup-java](https://github.com/actions/setup-java) only.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
           java-version: adopt@1.8
 
       - name: Cache sbt
-        uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


xRuiAlves (approved):
Cool beans! Looks good 👍

PR #382, opened 2022-11-17
Replace actions/cache with coursier/cache-action

jcazevedo:
Using coursier/cache-action directly removes the need to specify the cache directories. The action takes care of that for us.


xRuiAlves (approved):
Cool beans 👍

PR #380, opened 2022-11-16
Update AWS SDK from `1.x.x` to `2.x.x`

xRuiAlves:
Updating the underlying aws sdk library from version `1.x.x` to `2.x.x` in `apso-aws` and `apso-io`.

There are quite a few breaking changes, and I'm thus unsure how we want to approach versioning if this release goes forward!

PR #385, opened 2022-11-21
Add asdf's .tool-versions to .gitignore

jcazevedo:
Cherry-picked from #380. I was about to do a similar change in a different branch, so we might as well get this in separately.


xRuiAlves (approved):
Cool cool 👍

PR #386, opened 2022-11-21
Lowercase example names in `LocalFileDescriptorSpec`

jcazevedo:
It's not customary for examples to start with an upper case. This pull request makes the titles in `LocalFileDescriptiorSpec` consistent with the rest of the examples in the module.


xRuiAlves (approved):
Cool beans, thanks for updating! 👍

PR #432, opened 2023-02-23
Update `elasticsearch` to version `7.16.3`

xRuiAlves:


version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "0.18.7-SNAPSHOT"
+ThisBuild / version := "0.18.7"
```
dimamo5, line 1: I don't think we need to change the version here. This will be automatically updated with the release of a new version

dimamo5 (changes requested):

version.sbt
xRuiAlves: @dimamo5 yeap, you're right! Reverted the version updating commit, should be good now 👍

dimamo5 (changes requested):

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@
 
 ### Security
 
+## [0.18.7] - 2023-02-24
+
+### Changed
+- Update elasticsearch-related dependencies to `7.13.x` ([#432](https://github.com/adzerk/apso/pull/432)).
```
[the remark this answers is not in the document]
xRuiAlves: Oops, my bad! Fixed 👍

dimamo5 (dismissed):
Thanks

dimamo5 (approved):

PR #449, opened 2023-03-17
Add `SFTP` file descriptor credentials

xRuiAlves:
It's a bit awkward to work with these `(String, String, Either[...])` tuples from a consumer perspective. Adding a case class around it!

### DIFF
--- a/io/src/main/scala/com/velocidi/apso/io/SftpFileDescriptor.scala
+++ b/io/src/main/scala/com/velocidi/apso/io/SftpFileDescriptor.scala
@@ -239,7 +239,7 @@ object SftpFileDescriptor {
       ssh.disconnect()
     }
   }
-  
+
   case class SftpFileDescriptorCredentials(host: String, port: String, auth: Either[Identity, String])
 
   implicit def sftpClientToSFTPClient(c: SftpClient): SFTPClient = c.sftpClient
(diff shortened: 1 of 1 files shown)


io/src/main/scala/com/velocidi/apso/io/SftpFileDescriptor.scala
```diff
@@ -240,6 +240,8 @@ object SftpFileDescriptor {
     }
   }
 
+  case class SftpFileDescriptorCredentials(host: String, port: String, auth: Either[Identity, String])
```
jcazevedo, line 243: Since this is already scoped under the `SftpFileDescriptor` object, I think we can remove the prefix and just go with `Credentials`. We can use `SftpFileDescriptor.Credentials` for extra clarity.

jcazevedo (changes requested):
I just commented on a small naming issue but otherwise this looks good to me.

io/src/main/scala/com/velocidi/apso/io/SftpFileDescriptor.scala
xRuiAlves: Agreed! Updated 👍

jcazevedo (approved):
Thanks!

PR #470, opened 2023-04-27
Update Changelog in preparation for the 0.18.8 release

jcazevedo:
I have already set up a date because I plan on releasing this today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+<!--
+Use the following schema when setting up the Changelog for a new release. Remove empty sections before publishing.
+
 ## [Unreleased]
 
 ### Added
@@ -13,6 +16,7 @@
 ### Fixed
 
 ### Security
+-->
 
 ## [0.18.7] - 2023-02-28
 
@@ -60,7 +64,6 @@ This is the first version with a Changelog, albeit not being the first version o
 - Update Scala, dependencies and plugins versions ([#142](https://github.com/velocidi/apso/pull/142)).
 
 [0.18.1]: https://github.com/velocidi/apso/compare/v0.18.0...v0.18.1
-[Unreleased]: https://github.com/velocidi/apso/compare/v0.18.4...HEAD
 
 ***
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,47 @@ Use the following schema when setting up the Changelog for a new release. Remove
 ### Security
 -->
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,6 +16,48 @@
 ### Fixed
 
 ### Security
+-->
+
+## [0.18.8] - 2023-04-27
+
+This release includes several dependency updates, some updates to make existing APIs more friendlier and a change to a
```
xRuiAlves, line 23: Grammar nit:

```suggestion
This release includes several dependency updates, some updates to make existing APIs friendlier and a change to a
```

or even

```suggestion
This release includes several dependency updates, some updates to make existing APIs more friendly and a change to a
```

CHANGELOG.md
```diff
@@ -13,6 +16,48 @@
...
+```
+
+We have also promoted the SFTP file descriptor credentials from a tuple to a dedicated data type:
+
+```scala
+case class Credentials(host: String, port: String, auth: Either[Identity, String])
+```
+
+Additionally, the `Decoder` for [Squants](https://github.com/typelevel/squants)'s `Currency` was updated to become case
+insensitive, meaning that, for example, both `"usd"` and `"USD"` now decode to the same `USD` currency.
```
xRuiAlves, line 46: Perhaps a bit hard on the eye, too many commas. I would suggest something like:

```suggestion
insensitive. This means that, for example, both `"usd"` and `"USD"` now decode to the same `USD` currency.
```

xRuiAlves (approved):
Looks good, thanks!

Just left a couple minor comments

xRuiAlves:
These `SimpleLruCacheSpec` tests seem to be failing transiently quite often, unfortunately :/
Had the plan (on the back of my mind) to eventually try and do something about it, but haven't found the time to be honest 😅

CHANGELOG.md
jcazevedo: Yes, of course. At some point I had `more friendly` and wanted to go with just `friendlier`.

CHANGELOG.md
jcazevedo: Yes. I agree.

xRuiAlves (approved):
Cool beans!

jcazevedo:
> These `SimpleLruCacheSpec` tests seem to be failing transiently quite often, unfortunately :/ Had the plan (on the back of my mind) to eventually try and do something about it, but haven't found the time to be honest 😅

Yes. They are becoming quite annoying. I also have an item to address them in my todo list.

PR #485, opened 2023-06-04
Remove explicit dependency on the `organize-imports` Scalafix rule

jcazevedo:
This is a follow-up to #483. Since [Scalafix 0.11.0](https://github.com/scalacenter/scalafix/releases/tag/v0.11.0), `OrganizeImports` is a built-in rule.


beatriz (approved):
LGTM, thanks!

PR #543, opened 2023-09-17
Don't fail fast in CI matrix build

jcazevedo:
Lets the build of each Scala version run to its full extent, even if the other one fails.


xRuiAlves (approved):
Sounds good, thanks!

PR #549, opened 2023-10-04
Drop apso-log

thyandrecardoso:
Because `scalalogging` already provides the features present in `apso-log` it does not make much sense to have a module replicating them here. This PR remove `apso-log` module.


jcazevedo (approved):
Thanks!

PR #583, opened 2023-11-28
Upgrade Unirest dependency to 4.2.0

rafaavc:
Updated the Unirest dependency due to a concurrency issue that affected the version being used (https://github.com/Kong/unirest-java/issues/92), and that was showing up when using `W` in another project.

No

No

### How has this been tested?

Tested this manually.

### DIFF
--- a/core/build.sbt
+++ b/core/build.sbt
@@ -1,9 +1,6 @@
 import Dependencies._
 
 libraryDependencies ++= Seq(
-  ApacheHttpAsyncClient,
-  ApacheHttpClient,
-  ApacheHttpCore,
   CirceCore,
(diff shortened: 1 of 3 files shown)


jcazevedo (approved):
Thanks! 

One thing to note is that Unirest 4 requires at least Java 11 and we're currently publishing Apso (and running CI) in Java 8. This wasn't noticeable in CI because we don't use `W` in Apso's tests. We could keep publishing Apso using Java 8 and let downstream users deal with the incompatibilities (for those still in Java 8), but I think it might be safer to just embrace Java 11.

jcazevedo:
We can handle that in a separate PR.

PR #611, opened 2024-02-13
Fix compilation for Java 19+

JD557:
Addresses the following (fatal) deprecation warnings, which made it so that Apso would not compile on Java 19 and above.
- Java 19 deprecated [`Thread#getId`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Thread.html#getId()) in favor of the new [`Thread#threadId()`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Thread.html#threadId()).
- Java 20 deprecated the `URL` constructor in favor of `new URI(...).toURL` (See https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html#constructor-deprecation).

No

No

### How has this been tested?

By compiling the whole project using OpenJDK 21.0.2 and running the unit tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,31 +25,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
-      - name: Setup up JDK (Java 11)
-        id: setup-java-11
-        if: matrix.java == '11'
+      - name: Setup up JDK
         uses: actions/setup-java@v4
         with:
           distribution: adopt
-          java-version: 11
-          cache: sbt
-
-      - name: Setup up JDK (Java 17)
-        id: setup-java-17
-        if: matrix.java == '17'
-        uses: actions/setup-java@v4
-        with:
-          distribution: adopt
-          java-version: 17
-          cache: sbt
-
-      - name: Setup up JDK (Java 21)
-        id: setup-java-21
-        if: matrix.java == '21'
-        uses: actions/setup-java@v4
-        with:
-          distribution: adopt
-          java-version: 21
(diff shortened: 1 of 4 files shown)


JD557:
Added multiple Java versions to the build matrix.
However, this required me to change the test names, so this will require a branch protection bypass/update to be merged.

.github/workflows/ci.yml
```diff
@@ -24,11 +25,32 @@ jobs:
...
           java-version: 11
+          cache: sbt
+
+      - name: Setup up JDK (Java 17)
+        id: setup-java-17
+        if: matrix.java == '17'
+        uses: actions/setup-java@v4
+        with:
+          distribution: adopt
+          java-version: 17
```
jcazevedo, line 43: Can we refer to `matrix.java` here rather than having a step for each Java version?

JD557: I don't think so. I tried that in f7e15d717ef9aa7627f64576d5aab8d6eb0401bd and the CI failed to parse it (I force pushed over it to avoid a dirty history).

So I copied this approach from cats (not sure why they need all those `sbt +update`): https://github.com/typelevel/cats/blob/4f15360e99c27338859d7f0e413d2a38b8ca1f97/.github/workflows/ci.yml#L59-L96

jcazevedo (approved):
Nice. Thanks!

jcazevedo:
> However, this required me to change the test names, so this will require a branch protection bypass/update to be merged.

I changed the branch protection rules to refer to the new checks.

PR #612, opened 2024-02-14
Small improvement to JsonConvert performance

dimamo5:
Build a `Json` from a `JsonObject` when converting from a map. The internal data structure used is different and seems to be slightly more performant.

```
Benchmark        Mode  Cnt      Score     Error  Units
Bench.newBench  thrpt    5  11937.246 ± 213.424  ops/s
Bench.oldBench  thrpt    5  10711.585 ± 347.067  ops/s
```

No.

No.

### How has this been tested?

Running the test suite


dianaamfr (approved):
Looks great! Thanks! 👍🏼
This PR gave me the opportunity to interact with this repository for the first time and to experiment with JMH for benchmarking, so thank you!

PR #613, opened 2024-02-14
Improve retry logic with S3 client

pmscosta:

No.

No.

### How has this been tested?

I didn't test this but verified in production that with a https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/SdkClientException.html (which appears to inherit and keep the default value of true in the `isRetryable` method) we did not attempt to retry.


jcazevedo (approved):
Thanks!

PR #614, opened 2024-02-14
Mark Retry sleep as blocking

JD557:
The `Thread.sleep` in `Retry` was not being marked as blocking, so it was potentially blocking a thread. This fixes it so that other tasks can run while the future sleeps.

No

No

### How has this been tested?
This can be tested with the following snippet:
```scala
import com.velocidi.apso.Retry
import scala.concurrent.Future
import scala.concurrent.duration._
import scala.concurrent.ExecutionContext.Implicits.global

def badFuture(x: Int): Future[Unit] = Future {println(x); throw new Exception("Fail")}

(0 to 128).foreach(x => Retry.retryFuture(3, 1.second)(badFuture(x)))
```

The new logic will finish in ~4 seconds (as expected), while the old one will block for a while (unless your computer has more than 128 cores)

### DIFF
--- a/core/src/main/scala/com/velocidi/apso/Retry.scala
+++ b/core/src/main/scala/com/velocidi/apso/Retry.scala
@@ -1,7 +1,7 @@
 package com.velocidi.apso
 
 import scala.concurrent.duration._
-import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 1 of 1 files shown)


queimadus (approved):
LGTM

PR #616, opened 2024-02-16
Add pekko subprojects

JD557:
Adds `pekko`, `pekko-http` and `elasticsearch-pekko` modules.

These are basically copies of the `akka`, `akka-http` and `elasticsearch` modules, but ported to Pekko.

As a caveat, the classes in the `elasticsearch-pekko` module are under `com.velocidi.apso.elasticsearch.pekko`, to avoid potential conflicts if someone includes both `elasticsearch` and `elasticsearch-pekko` projects (while undesirable, this will happen if someone includes the `apso` bundle).

This also removes the dependency on `akka-testkit-specs2`, as it turns out it was not used that much.

No

Yes, a small note was added to the README

### How has this been tested?

Just using the unit tests.


dimamo5 (approved):
LGTM! Thank you

PR #656, opened 2024-04-22
Update nodejs version used in GitHub actions

xRuiAlves:
We are using github actions versions that, underlying, are using NodeJS 16 (which is mark for deprecation!).
We should bump their versions to the ones using NodeJS 20.

No.

### Can this PR be safely merged to the target branch after approval?

Yes.

### How has this been tested?

Relying on CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Setup node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: '18.18.0'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -52,10 +52,10 @@ jobs:
...
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: '18.18.0'
```
queimadus, line 60: Do you know if this override still needed?

xRuiAlves: It is, they don't provide a default.
However, they support using an `lts` value for the `node-version`, so we could also go in that direction?

queimadus: `lts` lgtm for the purpose of this github action 👍

xRuiAlves: Updated ✅

xRuiAlves:
Oops, seems like it can't find `lts` 🤔

queimadus (approved):
Thanks!

PR #677, opened 2024-05-17
Allow specifying `max-connections` on S3 client

pmscosta:
As titled.

It's useful to be able to tweak this setting in scenarios where we are dealing with many concurrent connections.

No.

I don't think so. Or, at least, I couldn't find any mention to the existing `region` setting.

### How has this been tested?

Ad-hoc testing locally.
I thought about creating unit tests to check that we are propagating this value correctly. However, all the relevant things are private and we already have specs for our ConfigLoader. We could think of a spec that checks that we are able to maintain more than X open connections, but I'm not sure if that's worth the hassle.

### DIFF
--- a/aws/src/main/scala/com/velocidi/apso/aws/S3Bucket.scala
+++ b/aws/src/main/scala/com/velocidi/apso/aws/S3Bucket.scala
@@ -36,6 +36,7 @@ class S3Bucket(
 
   private[this] lazy val configPrefix = "aws.s3"
   private[this] lazy val region = Try(config.getString(configPrefix + ".region"))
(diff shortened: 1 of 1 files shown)


joaocsf (approved):
LGTM. Published the package locally and appears to be behaving as expected. Our CI is failing, but I believe it is unrelated.

pmscosta:
> LGTM. Published the package locally and appears to be behaving as expected. Our CI is failing, but I believe it is unrelated.

Appeared to be so. Triggered a new run and looks good 👍

jcazevedo (approved):
Thanks!

PR #692, opened 2024-06-17
Fix race condition in LruCache when handling exceptions

JD557:
Reorders the operations in the `LruCache` to avoid race conditions.

When the future failed, the other threads could still fetch the failed result before the cache entry was deleted (this can be simulated by adding a `blocking(Thread.sleep(1000))` between the instructions).

Now the cache entry is deleted first, leaving other threads free to create new futures, before marking the returned future as failed.

This was detected by flaky tests in https://github.com/adzerk/apso/pull/690/checks.

No

No

### How has this been tested?

Unit tests.

I also reproduced the previous issue by changing the code to:

```scala
promise.complete(value)
scala.concurrent.blocking(Thread.sleep(1000))
if (value.isFailure) store.remove(key, promise.future)
```

Doing the same thing with the new ordering does not cause any issue.

### DIFF
--- a/caching/src/test/scala/com/velocidi/apso/caching/ExpiringLruCacheSpec.scala
+++ b/caching/src/test/scala/com/velocidi/apso/caching/ExpiringLruCacheSpec.scala
@@ -85,8 +85,7 @@ class ExpiringLruCacheSpec(implicit ee: ExecutionEnv) extends Specification {
       cache(2)(Future.successful("B")) must beEqualTo("B").await
       cache(3)("C") must beEqualTo("C").await
       cache.store.toString === "{1=A, 2=B, 3=C}"
-      cache(4)("D")
-      Thread.sleep(10)
+      cache(4)("D") must beEqualTo("D").await
       cache.store.toString === "{2=B, 3=C, 4=D}"
       cache.size === 3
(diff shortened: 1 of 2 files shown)


caching/src/test/scala/com/velocidi/apso/caching/ExpiringLruCacheSpec.scala
```diff
@@ -85,8 +85,7 @@ class ExpiringLruCacheSpec(implicit ee: ExecutionEnv) extends Specification {
       cache(2)(Future.successful("B")) must beEqualTo("B").await
       cache(3)("C") must beEqualTo("C").await
       cache.store.toString === "{1=A, 2=B, 3=C}"
-      cache(4)("D")
-      Thread.sleep(10)
```
JD557, line 89: I noticed that this test is now a bit more flaky.

I do think that we wanted the `await` here, not `Thread.sleep`.

jcazevedo: Yes. The `await` is more appropriate.

jcazevedo (approved):
Nice! Thanks!

PR #727, opened 2024-09-10
Force empty line after top-level package statement

pmscosta:
We try to always keep an empty line after the package statement at the top of the file, however we don't have this enforced by our scalafmt configuration files. This PR changes that in a hackish way given the following (as stated in [here](https://scalameta.org/scalafmt/docs/configuration.html#newlines-around-package-or-template-body)): 
> the rules do not directly apply to package statements at the top of the source file; however, if this parameter is non-empty, there will be at least one blank line before the first non-package statement, and possibly more if the rules match that statement

No.

No.

### How has this been tested?

Tested locally by adding some empty lines in other places we want them and check that this does not remove them. Also ran a root `scalafixAll`.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -2,6 +2,9 @@ version = 3.8.3
 
 runner.dialect = scala213
 maxColumn = 120
+newlines.topLevelStatementBlankLines = [
+  { blanks: { after: 0 } }
+]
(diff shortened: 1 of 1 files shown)


jcazevedo (approved):
Thank you!

PR #745, opened 2024-10-12
Rename packages organization from `com.velocidi` to `com.kevel`

beatriz:
This changes the packages from the `com.velocidi` to `com.kevel` organization to reflect the organizational change. For now, it doesn't touch the publishing organization as I still need to figure out some things for that.

After making the change I wondered if we'd want to use `adzerk` instead to be consistent with the GitHub organization name, what do you think?

I am leaving this with the `do not merge label` for now, as I know we'd like to get a Pekko update without this impacting its release. I will also try to figure out the publish organization to try to get it all at once.

No.

Updated the `README` docs as well.

### How has this been tested?

Relying on CI to validate everything keeps working.


jcazevedo:
> After making the change I wondered if we'd want to use `adzerk` instead to be consistent with the GitHub organization name, what do you think?

I'd rather not move from one old name (`com.velocidi`) to another old one (`com.adzerk`).

hashing/src/main/java/com/velocidi/apso/hashing/MurmurHash3.java
```diff
@@ -21,7 +21,7 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
  */
 
-package com.velocidi.apso.hashing;
+package com.kevel.apso.hashing;
```
jcazevedo, line 24: Can we rollback the other changes in this file? I suspect some unwanted IDE shenanigans went unnoticed here.

jcazevedo (changes requested):
Thanks for getting the ball rolling on this!

Can we also rename the `**/com/velocidi/*` folders to `**/com/kevel/*`?

beatriz:
> Can we also rename the `**/com/velocidi/*` folders to `**/com/kevel/*`?

Yeah, totally missed that 😅 Done!

jcazevedo (approved):
Thank you!

PR #756, opened 2024-11-05
Add a `uri` method to `FileDescriptor`

jcazevedo:
Enables the `FileDescriptor` interface with a `uri` method that returns a `java.net.URI` with a reference to the file's location.

No.

No.

### How has this been tested?

New unit tests were added for the change.


pmscosta (approved):
Clean! Thanks.

PR #759, opened 2024-11-20
Make fromFullPaths tail recursive

dianaamfr:
This proposes making `fromFullPaths` tail recursive to avoid `StackOverflow` exceptions if the size of the `paths` parameter is too long.

No.

No.

### How has this been tested?

Existing and new unit tests.

### DIFF
--- a/circe/src/test/scala/com/kevel/apso/circe/ImplicitsSpec.scala
+++ b/circe/src/test/scala/com/kevel/apso/circe/ImplicitsSpec.scala
@@ -33,8 +33,11 @@ class ImplicitsSpec extends Specification {
       }
 
       "giving precedence to the last path value if duplicate paths exist" in {
-        val res = fromFullPaths(List("a" -> 1.asJson, "a" -> 2.asJson, "a" -> 3.asJson))
-        res mustEqual json"""{"a": 3}"""
+        val json1 = fromFullPaths(List("a" -> 1.asJson, "a" -> 2.asJson, "a" -> 3.asJson))
+        json1 mustEqual json"""{"a": 3}"""
+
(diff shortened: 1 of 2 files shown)


circe/src/test/scala/com/kevel/apso/circe/ImplicitsSpec.scala
```diff
@@ -26,6 +26,16 @@ class ImplicitsSpec extends Specification {
...
       }
+
+      "not throw a StackOverflowError for large lists of paths" in {
+        val paths = (1 to 10000).map(i => (s"a.v$i", i.asJson)).toList
+        fromFullPaths(paths, ".") must not(throwAn[StackOverflowError])
+      }
+
+      "giving precedence to the last path value if duplicate paths exist" in {
+        val res = fromFullPaths(List("a" -> 1.asJson, "a" -> 2.asJson, "a" -> 3.asJson))
+        res mustEqual json"""{"a": 3}"""
```
jcazevedo, line 37: Since we're testing this, can we also add an example in which there are duplicate paths at different depths? Something like the following?

```suggestion
        val json1 = fromFullPaths(List("a" -> 1.asJson, "a" -> 2.asJson, "a" -> 3.asJson))
        json1 mustEqual json"""{"a": 3}"""

        val json2 = fromFullPaths(List("a.b.c" -> 1.asJson, "a.b" -> 2.asJson, "a" -> 3.asJson))
        json2 mustEqual json"""{"a": 3}"""
```

jcazevedo (approved):
Thank you! Looks good!

jcazevedo (approved):
Thank you!

PR #803, opened 2025-03-22
Have a single build.sbt file

jcazevedo:
This pull request proposes starting to have all the subproject build definitions under the root build.sbt file. This can be seen as a matter of personal style, but I feel like it helps with not jumping around the root build.sbt and the subproject's build.sbt when trying to find where something is defined. It is also the [style choice recommended by sbt](https://www.scala-sbt.org/1.x/docs/Multi-Project.html#Appendix%3A+Subproject+build+definition+files):

> We recommend putting all project declarations and settings in the root build.sbt file in order to keep all build definition under a single file. However, it’s up to you.

No.

No.

### How has this been tested?


bdmendes (approved):
Looks good to me!

PR #804, opened 2025-03-22
Remove shapeless as a dependency

jcazevedo:
We were not actively using it.

This depends on #803.

No.

### How has this been tested?


bdmendes (approved):
Thanks!

PR #805, opened 2025-03-22
Drop Akka modules

jcazevedo:
Following [Akka's licence change](https://akka.io/blog/why-we-are-changing-the-license-for-akka), we recommend downstream users to switch to [Pekko](https://pekko.apache.org/).

This depends on #803.

### How has this been tested?

I made sure we continue to be able to build the project successfully, and that the akka, akka-http and elasticsearch modules are no longer being published as part of Apso.


bdmendes (approved):
Thanks!

PR #807, opened 2025-03-22
Start cross-compiling most modules to Scala 3

jcazevedo:
This pull request enables cross-compiling most of Apso's modules to Scala 3. The modules that are still missing due to unmet dependencies are apso-elasticsearch and apso-io.

This depends on #806 and #804.

Yes. A mention to modules available for Scala 3 was added to the README.

### How has this been tested?

I made sure we continue to be able to build the project successfully and that we're able to publish modules for Scala 3 locally.


bdmendes (approved):
Great. Thank you!

PR #808, opened 2025-03-24
Start cross-compiling apso-io to Scala 3

jcazevedo:
This enables cross-compiling apso-io to Scala 3. [scala-pool](https://github.com/andresilva/scala-pool) is not available for Scala 3, but we can use the Scala 2.13 version.

This depends on #807.

No.

### How has this been tested?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -139,7 +139,7 @@ lazy val hashing = module(project, "hashing")
 lazy val io = module(project, "io")
   .dependsOn(aws, testkit % Test)
   .settings(
-    crossScalaVersions := List(Versions.Scala212, Versions.Scala213),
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -139,7 +139,7 @@ lazy val hashing = module(project, "hashing")
 lazy val io = module(project, "io")
   .dependsOn(aws, testkit % Test)
   .settings(
-    crossScalaVersions := List(Versions.Scala212, Versions.Scala213),
+    crossScalaVersions := List(Versions.Scala212, Versions.Scala213, Versions.Scala3),
```
bdmendes, line 142: As discussed offline, we can probably [ditch 2.12](https://github.com/adzerk/apso/issues/812), if you'd like to opportunistically handle that in another PR.

bdmendes (approved):

PR #809, opened 2025-03-24
Start cross-compiling apso-elasticsearch to Scala 3

jcazevedo:
This enables cross-compiling apso-elasticsearch to Scala 3. We had to cross-version elastic4s from Scala 2.13 and silence some deprecations, but I think it's working.

This depends on #807.

No.

### How has this been tested?


jcazevedo:
We're probably going to go a different route with apso-elasticsearch.

PR #810, opened 2025-03-25
Remove Guava option from apso-caching

jcazevedo:
The Guava implementation has been removed from ScalaCache in https://github.com/cb372/scalacache/pull/362. This follows along and removes it from apso-caching.

This depends on #803.

Yes. The documentation was updated accordingly.

### How has this been tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ To use it in an existing SBT project, add the following dependency to your `buil
 libraryDependencies += "com.kevel" %% "apso-caching" % "0.21.0"
 ```
 
-Apso provides utilities to simplify the caching of method calls, with [ScalaCache](https://cb372.github.io/scalacache/) and using either `Guava` or `Caffeine` as underlying cache implementations.
+Apso provides utilities to simplify the caching of method calls, with [ScalaCache](https://cb372.github.io/scalacache/) and using either `Caffeine` as the underlying cache implementation.
 
 These utilities are provided as `cached()` and `cachedF()` extension methods over all `FunctionN[]` types:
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -355,7 +355,7 @@ To use it in an existing SBT project, add the following dependency to your `buil
 libraryDependencies += "com.kevel" %% "apso-caching" % "0.21.0"
 ```
 
-Apso provides utilities to simplify the caching of method calls, with [ScalaCache](https://cb372.github.io/scalacache/) and using either `Guava` or `Caffeine` as underlying cache implementations.
+Apso provides utilities to simplify the caching of method calls, with [ScalaCache](https://cb372.github.io/scalacache/) and using either `Caffeine` as the underlying cache implementation.
```
bdmendes, line 358: Nit: `either` is not needed anymore.

bdmendes (commented):
Perhaps outside the scope of this PR, but should we provide a "default" API now? As a user, I'd probably find a `cachedF(duration)` API more pleasing than `cachedFImpl(...)` (what we have now), even more now that only 1 implementation is available.

jcazevedo:
> Perhaps outside the scope of this PR, but should we provide a "default" API now? As a user, I'd probably find a `cachedF(duration)` API more pleasing than `cachedFImpl(...)` (what we have now), even more now that only 1 implementation is available.

Yes. I'd rather tackle that in a different PR, because we're currently unsure of how the apso-caching API will look like in the migration to Scala 3.

bdmendes (approved):

PR #811, opened 2025-03-25
Remove apso-elasticsearch

jcazevedo:
This pull request proposes removing the apso-elasticsearch module. Its usefulness is currently debatable since it's pinned to a specific Elasticsearch version.

### Does this change relate to existing issues or pull requests?

No.

### Does this change require an update to the documentation?

No. Apparently there were no mentions to apso-elasticsearch in the documentation.

### How has this been tested?

I made sure we continued to be able to build the project successfully.

PR #816, opened 2025-03-27
Remove Scala Steward pins related to Elasticsearch dependencies

jcazevedo:
As a follow-up to the removal of apso-elasticsearch, we no longer need to pin these updates.

This is a follow-up to #811.

No.

### How has this been tested?

I made sure we weren't depending on any artifact mentioned in the removed list.


dimamo5 (approved):
Thank you!

PR #819, opened 2025-04-09
Avoid fetching all prefixes when listing S3 files under a directory

jcazevedo:
The `list` implementation of `S3FileDescriptor` was using the `listS3WithPrefix` method, which would list all files in a bucket matching a given prefix. This would be very costly if inner "folders" would contain a lot of files. The `ListObjectsRequest` object allows one to specify a delimiter, which speeds up the operation considerably when we're only interested in files at the next level (which is the case).

No.

No.

### How has this been tested?

This was tested ad-hoc.


joaocsf (approved):
Thanks! I validated these changes on a s3 folder with nested folders and files and it behaved as expected.

PR #813, opened 2025-03-25
Drop support for Scala 2.12

jcazevedo:
Recent Scala versions have been deprecating some Scala 2 syntax, and there doesn't seem to be a high demand for a Scala 2.12 version of Apso downstream. As such, this PR proposes removing support for Scala 2.12 altogether.

### Does this change relate to existing issues or pull requests?

Resolves #812.

### Does this change require an update to the documentation?

Yes. The README was updated accordingly.

### How has this been tested?

I made sure we continue to be able to build the project successfully.

PR #824, opened 2025-04-15
Remove `scala-collection-compat` dependency listing

bdmendes:
As titled; for hygiene although with no practical change. We forgot to do this in https://github.com/adzerk/apso/pull/813.

https://github.com/adzerk/apso/pull/813

No.

### How has this been tested?

CI.


jcazevedo (approved):
Nice catch! Thanks!

PR #826, opened 2025-04-16
Avoid specifying a default value for the maximum cache size

jcazevedo:
It's preferable to let downstream users decide on what maximum cache size to use.

This is a follow-up to #821.

### How has this been tested?

I'm relying on existing tests.


queimadus (approved):
Thanks!

bdmendes (approved):
Thanks!

PR #827, opened 2025-04-16
Start cross-compiling apso-caching to Scala 3

jcazevedo:
apso-caching was the only Apso module without a Scala 3 version. After #821, we're able to start cross-compiling it to Scala 3.

This is a follow-up to #821.

### How has this been tested?

I'm relying on a successful build of the whole project.


caching/src/main/scala-3/com/kevel/apso/caching/SyncEvidence.scala
```diff
@@ -0,0 +1,15 @@
...
+
+/** An evidence that an `A` is not a `Future[_]`.
+  *
+  * @tparam A
+  *   the tagged type.
+  */
+private[caching] trait SyncEvidence[A]
+
+private[caching] object SyncEvidence {
+  implicit def ev[A](implicit notFuture: NotGiven[A <:< Future[_]]): SyncEvidence[A] = new SyncEvidence[A] {}
```
bdmendes, line 14: Awesome!

bdmendes (approved):
Thank you!

PR #832, opened 2025-04-24
Fix non-determinism in CachedFunctionExtrasSpec failure eviction

bdmendes:
As titled. The fact that failing async computations will be evicted _eventually_ from cache is [already documented](https://github.com/adzerk/apso/blob/master/caching/src/main/boilerplate/com/kevel/apso/caching/CachedFunctionsExtras.scala.template#L34), but was not taken into consideration in this test.

Improves spec introduced in #821.

No.

### How has this been tested?

Ran the test a bunch of times, seems to not fail anymore.


queimadus (approved):
Thanks

Issue #815, opened 2025-03-26
Investigate how to migrate `apso-caching` to Scala 3


bdmendes:
`scala-cache` support for Scala 3 seems pretty stale. We should probably make our API agnostic of the underlying lib and/or provide another implementation.

PR #833, opened 2025-04-29
Add `implementation` method to `apso-caching` config class

bdmendes:
Adds a useful `implementation` method to `apso-caching` that exposes a raw `Scaffeine` implementation with a matching configuration. This will make advanced downstream usages simpler and more consistent.

This should not break binary compatibility.

Related to #815.

Yes, it was updated accordingly.

### How has this been tested?

New spec.


queimadus (approved):
Thanks

PR #837, opened 2025-04-30
Remove nil crossScalaVersions for root project

bdmendes:
Now that all Apso modules are Scala 3-compatible, I believe this should be fine.

No.

No.

### How has this been tested?

CI.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,7 +5,8 @@ import xerial.sbt.Sonatype.sonatypeCentralHost
 
 ThisBuild / organization := "com.kevel"
 
-ThisBuild / scalaVersion := Versions.Scala213
+ThisBuild / scalaVersion       := Versions.Scala213
+ThisBuild / crossScalaVersions := List(Versions.Scala213, Versions.Scala3)
 
 val javaVersion = "11"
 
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -5,9 +5,7 @@ import xerial.sbt.Sonatype.sonatypeCentralHost
 
 ThisBuild / organization := "com.kevel"
 
-// Setting no cross build for the aggregating root project so that we can have proper per project exclusions.
-ThisBuild / crossScalaVersions := Nil
```
jcazevedo, line 9: I suppose we can move the `crossScalaVersions` from `commonSettings` to the `ThisBuild` scope here, so that we have the configuration for the Scala versions in all projects close together.

jcazevedo (changes requested):

build.sbt
bdmendes: See [e9ebdc2](https://github.com/adzerk/apso/pull/837/commits/e9ebdc2973521c1604852ae598c4bc8da5437a95).

jcazevedo (approved):
Thank you!

PR #835, opened 2025-04-30
Add `Scaffeine` api mappings

bdmendes:
As titled; this will make `sbt:doc` happy.

### Does this change relate to existing issues or pull requests?

Yes, this should have been done in #833. I'll add CI for `sbt:doc` next.

### Does this change require an update to the documentation?

No.

### How has this been tested?

`sbt +doc`.

PR #839, opened 2025-04-30
Amend changelog for 0.22.1

bdmendes:
0.22.1 got delayed a bit due to a problem while releasing that got fixed in #835. Since it is not live yet, this updates the changelog to mention everything that is actually going to be included in the release.

Amend of #834.

No.

### How has this been tested?

N/A.


jcazevedo (approved):
Thank you!

Issue #823, opened 2025-04-11
Remove `Joda` dependency


bdmendes:
Since Java 8, `java.time` is the recommended way.

PR #849, opened 2025-05-29
Remove `nscala-time` dependency

beatriz:
We don't make much use of the dependency and removing it avoids polluting downstream projects using `apso-time`. I also added a `to` method to `DateTime` similar to `LocalDate`'s, as that method is provided by `nscala-time`, and this allows downstream dependencies to have access to this when just using `joda`.

Is a first step towards #823 and to help downstream projects migrate more incrementally.

Yes, updated the README accordingly and added a changelog entry for this.

### How has this been tested?

Unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,21 @@ Use the following schema when setting up the Changelog for a new release. Remove
 ### Security
 -->
 
+## [Unreleased]
+
+### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+- Dependency on `nscala-time`, using only `joda.time` methods instead ([#849](https://github.com/adzerk/apso/pull/849)).
+
+### Fixed
+
+### Security
+
 ## [0.22.1] - 2025-04-30
 
 This release reintroduces the `config.Cache.implementation` method in `apso-caching`, now typed in both the cache key
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,21 +18,6 @@ Use the following schema when setting up the Changelog for a new release. Remove
 ### Security
 -->
 
-## [Unreleased]
-
-### Added
-
-### Changed
-
-### Deprecated
-
-### Removed
-- Dependency on `nscala-time`, using only `joda.time` methods instead ([#849](https://github.com/adzerk/apso/pull/849)).
-
-### Fixed
-
-### Security
-
 ## [0.22.1] - 2025-04-30
 
 This release reintroduces the `config.Cache.implementation` method in `apso-caching`, now typed in both the cache key
--- a/time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
+++ b/time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
@@ -1,7 +1,6 @@
 package com.kevel.apso.time
 
-import com.github.nscala_time.time.Imports._
-import org.joda.time.ReadableInterval
+import org.joda.time.{DateTime, Interval, LocalDate, Period, ReadableInterval}
 
 /** A view of a time interval as an indexed sequence of `DateTimes`.
   */
@@ -29,17 +28,17 @@ trait IterableInterval extends IndexedSeq[DateTime] {
 case class SteppedInterval(interval: ReadableInterval, step: Period) extends IterableInterval {
 
   lazy val length: Int = {
-    var i = (interval.toDurationMillis / step.toDurationFrom(interval.getStart).millis).toInt
-    if (apply(i) < interval.getEnd) {
-      while (apply(i) <= interval.getEnd) { i += 1 }
+    var i = (interval.toDurationMillis / step.toDurationFrom(interval.getStart).getMillis).toInt
+    if (apply(i).isBefore(interval.getEnd)) {
+      while (!apply(i).isAfter(interval.getEnd)) { i += 1 }
       i
     } else {
-      while (apply(i) > interval.getEnd) { i -= 1 }
+      while (apply(i).isAfter(interval.getEnd)) { i -= 1 }
       i + 1
     } // FIXME more intelligent code for this?
   }
 
-  def apply(idx: Int) = interval.getStart + step.multipliedBy(idx)
+  def apply(idx: Int) = interval.getStart.plus(step.multipliedBy(idx))
 
   def by(newStep: Period) = new SteppedInterval(interval, newStep)
 }
@@ -71,8 +70,8 @@ object IterableInterval {
     */
   def apply(interval: ReadableInterval, step: Period, lastInclusive: Boolean = true): IterableInterval =
     if (lastInclusive) SteppedInterval(interval, step)
-    else if (interval.millis == 0) EmptySteppedInterval(step)
-    else SteppedInterval(interval.getStart to interval.getEnd - 1.millis, step)
+    else if (interval.toDuration.getMillis == 0) EmptySteppedInterval(step)
+    else SteppedInterval(new Interval(interval.getStart, interval.getEnd.minusMillis(1)), step)
 }
 
 /** A view of a time interval as an indexed sequence of `LocalDate`.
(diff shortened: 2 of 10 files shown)


time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
```diff
@@ -71,8 +70,8 @@ object IterableInterval {
     */
   def apply(interval: ReadableInterval, step: Period, lastInclusive: Boolean = true): IterableInterval =
     if (lastInclusive) SteppedInterval(interval, step)
-    else if (interval.millis == 0) EmptySteppedInterval(step)
-    else SteppedInterval(interval.getStart to interval.getEnd - 1.millis, step)
+    else if (interval.toDuration.getMillis == 0) EmptySteppedInterval(step)
+    else SteppedInterval(new Interval(interval.getStart, interval.getEnd.minusMillis(1)), step)
```
jcazevedo, line 74: `to` produces an interval that is inclusive of the end instant, but `Interval` is exclusive, so we shouldn't be subtracting a millisecond here.

Upon some closer inspection, I feel like the `lastInclusive` parameter of `IterableInterval.apply` is a dirty hack, since we can already convey that semantic using `Interval` alone.

What do you think of the following?

```diff
diff --git a/time/src/main/scala/com/kevel/apso/time/Implicits.scala b/time/src/main/scala/com/kevel/apso/time/Implicits.scala
index 97997bc5..9dafafea 100644
--- a/time/src/main/scala/com/kevel/apso/time/Implicits.scala
+++ b/time/src/main/scala/com/kevel/apso/time/Implicits.scala
@@ -46,7 +46,10 @@ object Implicits {
       */
     def to(d2: LocalDate) =
       LocalDateInterval(
-        IterableInterval(new Interval(d1.toDateTimeAtStartOfDay, d2.toDateTimeAtStartOfDay), Period.days(1), true)
+        IterableInterval(
+          new Interval(d1.toDateTimeAtStartOfDay, d2.toDateTimeAtStartOfDay.plusMillis(1)),
+          Period.days(1)
+        )
       )
 
     /** Returns an iterable interval starting at this `LocalDate` (inclusive) and ending at the given `LocalDate`
@@ -59,7 +62,7 @@ object Implicits {
       */
     def until(d2: LocalDate) =
       LocalDateInterval(
-        IterableInterval(new Interval(d1.toDateTimeAtStartOfDay, d2.toDateTimeAtStartOfDay), Period.days(1), false)
+        IterableInterval(new Interval(d1.toDateTimeAtStartOfDay, d2.toDateTimeAtStartOfDay), Period.days(1))
       )
   }
 
@@ -108,7 +111,7 @@ object Implicits {
       *   with a 1 day step.
       */
     def to(d2: DateTime): IterableInterval =
-      IterableInterval(new Interval(d1, d2), Period.days(1), true)
+      IterableInterval(new Interval(d1, d2.plusMillis(1)), Period.days(1))
 
     /** Returns an iterable interval starting at this `DateTime` (inclusive) and ending at the given `DateTime`
       * (exclusive), with a 1 day step.
@@ -118,7 +121,8 @@ object Implicits {
       *   an iterable interval starting at this `DateTime` (inclusive) and ending at the given `DateTime` (exclusive),
       *   with a 1 day step.
       */
-    def until(d2: DateTime): IterableInterval = IterableInterval(new Interval(d1, d2), Period.days(1), false)
+    def until(d2: DateTime): IterableInterval =
+      IterableInterval(new Interval(d1, d2), Period.days(1))
   }
 
   /** Implicit class that provides new methods for `ReadableIntervals`.
@@ -138,9 +142,7 @@ object Implicits {
       if (n == 0) Seq.empty
       else {
         val q = (interval.toDuration.getMillis / n).toInt
-        (0 until n).map { i =>
-          new Interval(interval.getStart.plus(q * i + i), (interval.getStart.plus(q * (i + 1)).plus(i)))
-        }
+        (0 until n).map(i => new Interval(interval.getStart.plus(q * i), (interval.getStart.plus(q * (i + 1)))))
       }
     }
   }
diff --git a/time/src/main/scala/com/kevel/apso/time/IterableInterval.scala b/time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
index dd72f544..4ff7137e 100644
--- a/time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
+++ b/time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
@@ -1,6 +1,6 @@
 package com.kevel.apso.time
 
-import org.joda.time.{DateTime, Interval, LocalDate, Period, ReadableInterval}
+import org.joda.time.{DateTime, LocalDate, Period, ReadableInterval}
 
 /** A view of a time interval as an indexed sequence of `DateTimes`.
   */
@@ -30,10 +30,10 @@ case class SteppedInterval(interval: ReadableInterval, step: Period) extends Ite
   lazy val length: Int = {
     var i = (interval.toDurationMillis / step.toDurationFrom(interval.getStart).getMillis).toInt
     if (apply(i).isBefore(interval.getEnd)) {
-      while (!apply(i).isAfter(interval.getEnd)) { i += 1 }
+      while (apply(i).isBefore(interval.getEnd)) { i += 1 }
       i
     } else {
-      while (apply(i).isAfter(interval.getEnd)) { i -= 1 }
+      while (!apply(i).isBefore(interval.getEnd)) { i -= 1 }
       i + 1
     } // FIXME more intelligent code for this?
   }
@@ -63,15 +63,12 @@ object IterableInterval {
     *   the `ReadableInterval` to view as an indexed sequence
     * @param step
     *   the period of time between consecutive `DateTimes`
-    * @param lastInclusive
-    *   `true` if the upper bound of the interval is to be included in the sequence
     * @return
     *   an iterable time interval with the given step.
     */
-  def apply(interval: ReadableInterval, step: Period, lastInclusive: Boolean = true): IterableInterval =
-    if (lastInclusive) SteppedInterval(interval, step)
-    else if (interval.toDuration.getMillis == 0) EmptySteppedInterval(step)
-    else SteppedInterval(new Interval(interval.getStart, interval.getEnd.minusMillis(1)), step)
+  def apply(interval: ReadableInterval, step: Period): IterableInterval =
+    if (interval.toDuration.getMillis == 0) EmptySteppedInterval(step)
+    else SteppedInterval(interval, step)
 }
 
 /** A view of a time interval as an indexed sequence of `LocalDate`.
diff --git a/time/src/test/scala/com/kevel/apso/time/ImplicitsSpec.scala b/time/src/test/scala/com/kevel/apso/time/ImplicitsSpec.scala
index 1226d856..699c33e4 100644
--- a/time/src/test/scala/com/kevel/apso/time/ImplicitsSpec.scala
+++ b/time/src/test/scala/com/kevel/apso/time/ImplicitsSpec.scala
@@ -101,11 +101,11 @@ class ImplicitsSpec extends Specification {
 
   "An ApsoTimeInterval" should {
     "split itself into subsequences" in {
-      val interval = new Interval(0, 99)
+      val interval = new Interval(0, 100)
       interval.split(0) === Seq.empty
       interval.split(-1) must throwAn[IllegalArgumentException]
       interval.split(5).size === 5
-      interval.split(4) === Seq(new Interval(0, 24), new Interval(25, 49), new Interval(50, 74), new Interval(75, 99))
+      interval.split(4) === Seq(new Interval(0, 25), new Interval(25, 50), new Interval(50, 75), new Interval(75, 100))
     }
   }
 }
```

CHANGELOG.md
```diff
@@ -18,6 +18,21 @@ Use the following schema when setting up the Changelog for a new release. Remove
...
+## [Unreleased]
+
+### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+- Dependency on `nscala-time`, using only `joda.time` methods instead ([#849](https://github.com/adzerk/apso/pull/849)).
```
jcazevedo, line 30: It's OK to not eagerly update the Changelog if you don't feel like it. We usually run something like the following prior to a release to figure out which items to include.

```sh
gh pr list --limit 100 --state merged --search "base:master merged:>2025-04-07T10:09:18+0100" --json title,url,number --template '{{range .}}- {{.title}} ([#{{.number}}]({{.url}})).{{"\n"}}{{end}}' | tac
```

time/src/main/scala/com/kevel/apso/time/IterableInterval.scala
beatriz: Yeah, I totally missed that. And I agree with the `lastInclusive`, so I just applied your diff.

jcazevedo (approved):
Thanks a lot for this. Looks great!

PR #885, opened 2025-08-05
Add `StrictMatching` trait to `specs2_5` testkit

bdmendes:
In a quest for a more opinionated, simpler testing schema with `specs2`, we can use the new negation implicits introduced in `specs2_5`. The main purpose of this is to establish the use of `must` as the only syntax available.

No.

It can, if you think it is worth.

### How has this been tested?

I've included it in a downstream project and verified that the methods do not exist.

### DIFF
--- a/specs2_5/src/main/scala/com/kevel/apso/StrictMatching.scala
+++ b/specs2_5/src/main/scala/com/kevel/apso/StrictMatching.scala
@@ -1,0 +1,17 @@
+package com.kevel.apso
+
+import org.specs2.matcher.*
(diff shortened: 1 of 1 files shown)


beatriz:
I am not sure if this is something that belongs to this project, as it is super opinionated and would make more sense to be setup by each downstream user of `apso`..

bdmendes:
That's fair. This is completely opt-in, but it's indeed opinionated.

bdmendes:
Let's close this for now.

PR #911, opened 2025-10-07
Read the most recent version for the docs from the Changelog

jcazevedo:
This pull request proposes fetching the most recent version we refer to the in the README from the most recent header in the Changelog. We frequently forget to update the `VERSION` mdoc variable so I think this helps to keep things in sync.

No.

This is an update to the documentation.

### How has this been tested?

I manually executed `docs/mdoc` and made sure the `VERSION` variable was properly filled, resulting in the update to `README.md` included in this pull request.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -164,7 +164,10 @@ lazy val docs = (project in file("apso-docs"))
     mdocOut := (ThisBuild / baseDirectory).value,
 
     mdocVariables := Map(
-      "VERSION" -> "0.25.1" // This version should be set to the currently released version.
+      "VERSION" ->
+        // This is reading the most recent header from the CHANGELOG that looks like a version, assuming that's the most
+        // recent version to refer to in the docs.
+        IO.readLines(file("CHANGELOG.md")).flatMap("""^## \[([0-9.]+)\]""".r.findFirstMatchIn).head.group(1)
     ),
 
     publish / skip := true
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -164,7 +164,10 @@ lazy val docs = (project in file("apso-docs"))
     mdocOut := (ThisBuild / baseDirectory).value,
 
     mdocVariables := Map(
-      "VERSION" -> "0.25.1" // This version should be set to the currently released version.
+      "VERSION" ->
+        // This is reading the most recent header from the CHANGELOG that looks like a version, assuming that's the most
+        // recent version to refer to in the docs.
+        IO.readLines(file("CHANGELOG.md")).flatMap("""^## \[([0-9.]+)\]""".r.findFirstMatchIn).head.group(1)
```
bdmendes, line 170: I am fine with this approach. But as food for thought, could we reason about the returned value in comparison to what's in `version.sbt`? It would be nice to be able to catch an error like a wrong header in the md file.

jcazevedo: Maybe, but that's hard to solve in general because `version.sbt` holds the "current" version and this `VERSION` should include the last published version, so we would need some mechanism able to compare versions (assuming a scheme and snapshot versions). Since an update to this `VERSION` variable results in an update to the README, I think mistakes are easy to spot as is.

bdmendes (approved):

PR #1118, opened 2026-07-18
Make SyncEvidence a public class

bdmendes:
As of [Scala 3.8+](https://github.com/scala/scala3/pull/25367), and upcoming [Scala 2.13.19](https://github.com/scala/scala/pull/11215), resolving implicits from private companion objects is deprecated.

Opportunistically, I took the chance to make our `SyncEvidence` a simple class, as it has no behavior, and modernize the given in the Scala 3 version. The latter might raise binary binary compatibility issues as the signature is changed, but we don't promise stability in Apso. In any case, the `caching` module is typically used as a leaf in application code, so this should not matter too much.

No, but the Scala 3.9 LTS version is around the corner and downstream projects started firing the warning:

```scala
[warn] xx |  }).cachedSync(config.Cache(None, None))
[warn]    |   ^
[warn]    |Usage of implicit method ev defined in object SyncEvidence, which is not accessible here. In Scala 3.10, this implicit will no longer be found.
```

No.

### How has this been tested?

Existing spec.

### DIFF
--- a/modules/caching/src/main/scala-2/com/kevel/apso/caching/SyncEvidence.scala
+++ b/modules/caching/src/main/scala-2/com/kevel/apso/caching/SyncEvidence.scala
@@ -9,7 +9,7 @@ import scala.concurrent.Future
   * @tparam A
   *   the tagged type.
   */
-class SyncEvidence[A]
+class SyncEvidence[A] private[caching] ()
 
 object SyncEvidence extends LowPrioritySyncEvidence {
   implicit def ev1[A]: SyncEvidence[Future[A]] = new SyncEvidence[Future[A]]
--- a/modules/caching/src/main/scala-2/com/kevel/apso/caching/SyncEvidence.scala
+++ b/modules/caching/src/main/scala-2/com/kevel/apso/caching/SyncEvidence.scala
@@ -9,13 +9,13 @@ import scala.concurrent.Future
   * @tparam A
   *   the tagged type.
   */
-private[caching] trait SyncEvidence[A]
+class SyncEvidence[A]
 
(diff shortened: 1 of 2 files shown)


jcazevedo (changes requested):
> As of [Scala 3.8+](https://github.com/scala/scala3/pull/25367), and upcoming [Scala 2.13.19](https://github.com/scala/scala/pull/11215), resolving implicits from private companion objects is deprecated.

I don't think this is entirely accurate. What was deprecated was the access from non-accessible companion objects (but they could still be private).

Can't we keep `SyncEvidence` package-private and just make the companion object public? That way at least we avoid unintentional creations of `SyncEvidence` instances downstream.

bdmendes:
> What was deprecated was the access from non-accessible companion objects

Yes, *inaccessible* would be the more precise way to refer to it not being available in downstream projects. That was what I was referring to.

> Can't we keep SyncEvidence package-private and just make the companion object public

We can, and this warning is indeed only related to implicit search under companion objects, but as a general principle, for me it's quite odd for a public library method to require an object of a non-accessible class. In [035e3cc](https://github.com/adzerk/apso/pull/1118/commits/035e3cc554049a97285d7993cfa86b95ddb007f5) I am keeping the class public, but making its constructor private to address your concern. Let me know what you think about this.

jcazevedo (approved):
Thanks for this!

> In [035e3cc](https://github.com/adzerk/apso/pull/1118/commits/035e3cc554049a97285d7993cfa86b95ddb007f5) I am keeping the class public, but making its constructor private to address your concern. Let me know what you think about this.

That's reasonable!

PR #1120, opened 2026-07-18
Tighten `scalafmt` settings

bdmendes:
We have seen great results (for consistency) of the scalafmt rewrites introduced here in other projects. In parallel, this removes the `sbt1` override for the now sbt2 files.

No.

No.

### How has this been tested?

CI.


.scalafmt.conf
```diff
@@ -6,20 +6,11 @@ newlines.topLevelStatementBlankLines = [
   { blanks: { after: 0 } }
 ]
 
-fileOverride {
-  "glob:**/*.sbt" {
-    runner.dialect = sbt1
-    align.preset = most
-  }
-  "glob:**/project/*.scala" {
-    align.preset = most
```
jcazevedo, line 15: Would it be OK to keep the alignment in the build files? I find that it makes at least the dependency lists much easier to read.

bdmendes: Was planning to "hard reset" the scalafmt settings, but I agree that it looks better the old way. Please see [d12fd9d](https://github.com/adzerk/apso/pull/1120/commits/d12fd9dec00bd824e67fb2124943f464161b0301).

jcazevedo (approved):
Thank you!

PR #1128, opened 2026-07-20
Add move operation to `FileDescriptor` and refine base typing

bdmendes:
Adds a new move operation to `FileDescriptor` that behaves similarly to UNIX's `mv` utility.

- Moving directories will unsupported file systems (e.g. GCS or S3) will throw, as `upload` already does.
- Failing the operation, due to some I/O error, returns None.
- A successful move returns the new file descriptor.

Most notably, I am interested in using the `move` method for sftp.

Opportunistically, Self was moved up, requiring implementations to return their own type (which they already did), and using it for the `mvLocation` method up in the hierarchy.

No.

### How has this been tested?

Tested the local version via new unit tests.
Tested the sftp version against a local sftp server.

### DIFF
--- a/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
+++ b/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
@@ -210,7 +210,7 @@ trait FileDescriptor {
 
   /** Move the file to a new location in the same file system. Cross file system moves are not supported.
     *
-    * This operation may be more efficient than a manual [[upload]] followed by a [[delete]].
+    * This operation may be more efficient than a manual `upload` followed by a `delete`.
     *
     * @param pathString
     *   the new location, specified using unix's mv syntax related to the current directory.
--- a/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
+++ b/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
@@ -205,7 +205,9 @@ trait FileDescriptor {
     */
   protected final def mvLocation(pathString: String): Self = {
     val requestedDestination = parent().cd(pathString).asInstanceOf[Self]
-    if (requestedDestination.isDirectory) requestedDestination.child(name).asInstanceOf[Self] else requestedDestination
+    if (requestedDestination.exists && requestedDestination.isDirectory)
+      requestedDestination.child(name).asInstanceOf[Self]
+    else requestedDestination
   }
 
   /** Move the file to a new location in the same file system. Cross file system moves are not supported.
--- a/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
+++ b/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
@@ -10,6 +10,8 @@ import scala.io.Source
   */
 trait FileDescriptor {
 
+  type Self <: FileDescriptor
+
   /** Returns the unique identifier of this file in its location.
     * @return
     *   the unique identifier of this file in its location.
@@ -96,7 +98,7 @@ trait FileDescriptor {
     * @return
     *   a iterator of file descriptors
     */
-  def list: Iterator[FileDescriptor]
+  def list: Iterator[Self]
 
   /** Lists all files down the file hierarchy tree that whose relative path match the given prefix
     * @param prefix
@@ -104,7 +106,7 @@ trait FileDescriptor {
     * @return
     *   a iterator of file descriptors
     */
-  def listAllFilesWithPrefix(prefix: String): Iterator[FileDescriptor]
+  def listAllFilesWithPrefix(prefix: String): Iterator[Self]
 
   /** Returns the file descriptor `n` node up in the filesystem path.
     * @param n
@@ -112,7 +114,7 @@ trait FileDescriptor {
     * @return
     *   the file descriptor `n` node up in the filesystem path.
     */
-  def parent(n: Int = 1): FileDescriptor
+  def parent(n: Int = 1): Self
 
   /** Adds a new child node to the filesystem path.
     * @param name
@@ -120,7 +122,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def child(name: String): FileDescriptor
+  def child(name: String): Self
 
   /** Adds a new child node to the filesystem path.
     * @param name
@@ -128,7 +130,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def /(name: String): FileDescriptor = child(name)
+  def /(name: String): Self = child(name)
 
   /** Adds multiple new child nodes to the filesystem path.
     * @param names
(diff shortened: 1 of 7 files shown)


modules/io/src/main/scala/com/kevel/apso/io/LocalFileDescriptor.scala
```diff
@@ -188,6 +191,18 @@ case class LocalFileDescriptor(initialPath: String) extends FileDescriptor with
...
 
+  def move(pathString: String): Option[LocalFileDescriptor] = {
+    val destination = mvLocation(pathString)
+    if (destination.path == path) Some(this)
+    else
+      try {
+        Files.move(normalizedPath, destination.normalizedPath, StandardCopyOption.ATOMIC_MOVE)
+        Some(destination)
+      } catch {
+        case NonFatal(_) => None
```
afonsoafonsoafonso: Should we log a warning in these cases? Just for visibility.

modules/io/src/main/scala/com/kevel/apso/io/SftpFileDescriptor.scala
```diff
@@ -141,6 +141,13 @@ case class SftpFileDescriptor(
   def mkdirs(): Boolean =
     Try(sftp(_.mkdirs(path))).isSuccess
 
+  def move(pathString: String): Option[SftpFileDescriptor] = {
+    val destination = mvLocation(pathString)
+    if (destination.path == path) Some(this)
+    else if (destination.parent().mkdirs() && Try(sftp(_.rename(path, destination.path))).isSuccess) Some(destination)
```
afonsoafonsoafonso, line 147: Should we delete the created directory if the rename fails?

modules/io/src/main/scala/com/kevel/apso/io/S3FileDescriptor.scala
```diff
@@ -127,6 +128,15 @@ case class S3FileDescriptor(
...
 
+  def move(pathString: String): Option[S3FileDescriptor] = {
+    val destination = mvLocation(pathString)
+    if (destination.elements == elements) Some(this)
+    else {
+      val copied = Using.resource(stream())(input => destination.upload(input, Some(size)))
+      if (copied && delete()) Some(destination.copy(summary = None)) else None
+    }
+  }
+
```
afonsoafonsoafonso, line 139: Can't we take advantage of native server-side copy (we do mention _"may be more efficient than a manual upload followed by a delete"_)? Or is it not possible?

modules/io/src/main/scala/com/kevel/apso/io/SftpFileDescriptor.scala
```diff
@@ -141,6 +141,13 @@ case class SftpFileDescriptor(
   def mkdirs(): Boolean =
     Try(sftp(_.mkdirs(path))).isSuccess
 
+  def move(pathString: String): Option[SftpFileDescriptor] = {
+    val destination = mvLocation(pathString)
```
afonsoafonsoafonso, line 145: Should `mvLocation` here recover and return `None` upon a transient error?

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -193,6 +195,31 @@ trait FileDescriptor {
...
+  /** Move the file to a new location in the same file system. Cross file system moves are not supported.
+    *
+    * This operation may be more efficient than a manual `upload` followed by a `delete`.
+    *
+    * @param pathString
+    *   the new location, specified using unix's mv syntax related to the current directory.
+    * @return
+    *   the new file descriptor with the updated path if the move was successful, None otherwise.
+    */
+  def move(pathString: String): Option[Self]
```
afonsoafonsoafonso, line 222: Are the S3 and GCS's concrete implementations consistent with the Local/SFTP ones and between these ScalaDocs and the PR's description? I think the former currently only return `None` when upload/delete fails, with the underlying layer (`S3Bucket.stream`/`GCSBucket.push`) throwing upon IO errors.

Should we wrap these in some `NonFatal`/`Try` handling?

afonsoafonsoafonso, line 222: Additionally, would it make sense to extract the common logic between the `move` implementations to the base trait? We could have the concrete implementations implement a kind of `moveTo` with the backend-specific call and the rest could be extracted and kept in the trait, I believe.

afonsoafonsoafonso (changes requested):

modules/io/src/main/scala/com/kevel/apso/io/SftpFileDescriptor.scala
bdmendes: `mvLocation` should in principle not fail. It can, however, as it may fetch the remote to understand if the requested destination is a directory. We could do:
```scala
protected final def mvLocation(pathString: String): Option[Self] = {
  val requestedDestination = parent().cd(pathString).asInstanceOf[Self]
  for {
    isDirectory <- Try(requestedDestination.exists && requestedDestination.isDirectory).toOption
  } yield
    if (isDirectory) requestedDestination.child(name).asInstanceOf[Self]
    else requestedDestination
}
```
And act on it. But then, we should probably review whether `exists` and `isDirectory` should also return `Option` as they are the failure points. Given the current status quo – where exceptions may occur – I'd be inclined to keep this as is. Let me know what you think.

modules/io/src/main/scala/com/kevel/apso/io/S3FileDescriptor.scala
bdmendes: In GCS and S3, there is no move operation, so I don't think it is possible to make those more efficient than just a copy+delete. Locally and in sftp the implementation is specialized.

modules/io/src/main/scala/com/kevel/apso/io/SftpFileDescriptor.scala
bdmendes: We could, but it is difficult. We don't really know how many directories `mkDirs` created along the way. We could use some is empty + just created heuristic; please let me know if you have another suggestion.

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: We can resort to `Using.apply` instead of `Using.resource` for that. But again, I'd prefer, if we care to make these functions actually more transparent, to start from the leaves and make `upload` a fallible operation.

bdmendes: Tried to do that with `mvLocation`. Unfortunately the rest of the body is kind of specific, some do move, some do copy+delete, some create intermediate directories... The only clear "single point" candidates are S3 and GCS, but there is no common trait for them at the moment, and it's perhaps an over-abstraction to introduce one.

modules/io/src/main/scala/com/kevel/apso/io/LocalFileDescriptor.scala
bdmendes: Yes. Please see [6a282f2](https://github.com/adzerk/apso/pull/1128/commits/6a282f261271a4f3db22d5b209c945e70d2fa06a).

afonsoafonsoafonso (approved):
Lgtm! It's noticeable that this is a bit legacy-ish, I agree with what we discussed offline about not going the extra mile to improve it since we've been going more towards other solutions.

PR #1129, opened 2026-07-20
Remove unnecessary `FileDescriptor` overrides

bdmendes:
These operations are safe to perform at the base. There is no need to repeat them in implementations – they were sometimes slightly more efficient (e.g. not allocating intermediate wrappers while descending through child directories) but I don't think that is significant enough to maintain more code.

Opportunistic change after https://github.com/adzerk/apso/pull/1128.

No.

### How has this been tested?

I am relying on the soundness of the `parent` and `child` implementations.

### DIFF
--- a/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
+++ b/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
@@ -86,7 +86,7 @@ trait FileDescriptor {
     * @return
     *   an iterator with the lines of this file.
     */
-  final def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
+  def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
 
   /** Returns true if the fd points to a directory
     * @return
@@ -138,7 +138,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  final def children(names: String*): Self =
+  def children(names: String*): Self =
     names.foldLeft(this.asInstanceOf[Self])((acc, c) => acc.child(c).asInstanceOf[Self])
 
   /** Changes the path of the file descriptor using unix's cd syntax related to the current directory.
@@ -176,7 +176,7 @@ trait FileDescriptor {
     * @return
     *   a new file descriptor pointing to a sibling of the current file descriptor
     */
-  final def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
+  def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
 
   /** Returns true if the file pointed by the file descriptor exists
     * @return
--- a/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
+++ b/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
@@ -155,7 +155,7 @@ trait FileDescriptor {
     *   the new file descriptor with the updated path
     */
   final def cd(pathString: String): Self = {
-    pathString.split("/").toList.foldLeft(this.asInstanceOf[Self]) {
+    pathString.split("/").map(_.trim).toList.foldLeft(this.asInstanceOf[Self]) {
       case (acc, "." | "") => acc
       case (acc, "..")     => acc.parent().asInstanceOf[Self]
       case (acc, segment)  => acc.child(segment).asInstanceOf[Self]
--- a/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
+++ b/modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
@@ -86,7 +86,7 @@ trait FileDescriptor {
     * @return
     *   an iterator with the lines of this file.
     */
-  def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
+  final def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
 
   /** Returns true if the fd points to a directory
     * @return
@@ -130,7 +130,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def /(name: String): Self = child(name)
+  final def /(name: String): Self = child(name)
 
   /** Adds multiple new child nodes to the filesystem path.
     * @param names
@@ -138,7 +138,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def children(names: String*): Self =
+  final def children(names: String*): Self =
     names.foldLeft(this.asInstanceOf[Self])((acc, c) => acc.child(c).asInstanceOf[Self])
 
   /** Changes the path of the file descriptor using unix's cd syntax related to the current directory.
@@ -154,7 +154,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def cd(pathString: String): Self = {
+  final def cd(pathString: String): Self = {
     pathString.split("/").toList.foldLeft(this.asInstanceOf[Self]) {
       case (acc, "." | "") => acc
       case (acc, "..")     => acc.parent().asInstanceOf[Self]
@@ -168,7 +168,7 @@ trait FileDescriptor {
     * @return
     *   a new file descriptor pointing to a sibling of the current file descriptor
     */
-  def sibling(name: String): Self = sibling(_ => name)
+  final def sibling(name: String): Self = sibling(_ => name)
 
   /** Returns a new file descriptor pointing to a sibling of the current file descriptor
     * @param f
@@ -176,7 +176,7 @@ trait FileDescriptor {
     * @return
     *   a new file descriptor pointing to a sibling of the current file descriptor
     */
-  def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
+  final def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
 
   /** Returns true if the file pointed by the file descriptor exists
     * @return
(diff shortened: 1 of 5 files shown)


modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -86,7 +86,7 @@ trait FileDescriptor {
     * @return
     *   an iterator with the lines of this file.
     */
-  def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
+  final def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
```
jcazevedo: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def lines(): Iterator[String] = Source.fromInputStream(stream()).getLines()
```

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -130,15 +130,15 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def /(name: String): Self = child(name)
+  final def /(name: String): Self = child(name)
```
jcazevedo, line 133: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def /(name: String): Self = child(name)
```

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -130,15 +130,15 @@ trait FileDescriptor {
...
 
   /** Adds multiple new child nodes to the filesystem path.
     * @param names
     *   the node names to add.
     * @return
     *   the new file descriptor with the updated path
     */
-  def children(names: String*): Self =
+  final def children(names: String*): Self =
```
jcazevedo: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def children(names: String*): Self =
```

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -154,7 +154,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def cd(pathString: String): Self = {
+  final def cd(pathString: String): Self = {
```
jcazevedo: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def cd(pathString: String): Self = {
```

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -168,15 +168,15 @@ trait FileDescriptor {
     * @return
     *   a new file descriptor pointing to a sibling of the current file descriptor
     */
-  def sibling(name: String): Self = sibling(_ => name)
+  final def sibling(name: String): Self = sibling(_ => name)
```
jcazevedo, line 171: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def sibling(name: String): Self = sibling(_ => name)
```

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -168,15 +168,15 @@ trait FileDescriptor {
...
 
   /** Returns a new file descriptor pointing to a sibling of the current file descriptor
     * @param f
     *   a function that returns a new name from the current name of the file descriptor
     * @return
     *   a new file descriptor pointing to a sibling of the current file descriptor
     */
-  def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
+  final def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
```
jcazevedo: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def sibling(f: String => String): Self = parent().child(f(name)).asInstanceOf[Self]
```

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -154,7 +154,7 @@ trait FileDescriptor {
     * @return
     *   the new file descriptor with the updated path
     */
-  def cd(pathString: String): Self = {
+  final def cd(pathString: String): Self = {
     pathString.split("/").toList.foldLeft(this.asInstanceOf[Self]) {
```
jcazevedo: All the implementations of this were trimming the individual splits. I can't recall why. Maybe we were just being overcautious for malformed inputs. That being said, it might not be a bad idea to keep the trimming in the overarching implementation for the time being.

modules/io/src/main/scala/com/kevel/apso/io/RemoteFileDescriptor.scala
```diff
@@ -25,12 +25,9 @@ trait RemoteFileDescriptor { this: FileDescriptor =>
 
   protected def duplicate(elements: List[String]): Self
 
-  def parent(n: Int = 1): Self =
+  final def parent(n: Int = 1): Self =
```
jcazevedo, line 28: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def parent(n: Int = 1): Self =
```

modules/io/src/main/scala/com/kevel/apso/io/RemoteFileDescriptor.scala
```diff
@@ -25,12 +25,9 @@ trait RemoteFileDescriptor { this: FileDescriptor =>
...
     this.duplicate(elements = elements.dropRight(n))
 
-  def child(name: String): Self =
+  final def child(name: String): Self =
```
jcazevedo, line 31: I think it's still reasonable to give the option of implementations to redefine this, no?

```suggestion
  def child(name: String): Self =
```

jcazevedo (changes requested):
Thanks for this!

I think this change makes sense, but I would still keep open the opportunity for implementations to override specific methods (i.e. remove the `final`s), particularly with `FileDescriptor` not being sealed.

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: Please see [19f3ab3](https://github.com/adzerk/apso/pull/1129/commits/19f3ab3f8699001933876171715de50631c7b751).

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: Please see [19f3ab3](https://github.com/adzerk/apso/pull/1129/commits/19f3ab3f8699001933876171715de50631c7b751).

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: Brought back the trimming in [1549f5e](https://github.com/adzerk/apso/pull/1129/commits/1549f5ef8de07ec0ff9243391429e918b06ffe8c).

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: Please see [1549f5e](https://github.com/adzerk/apso/pull/1129/commits/1549f5ef8de07ec0ff9243391429e918b06ffe8c).

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: Please see [1549f5e](https://github.com/adzerk/apso/pull/1129/commits/1549f5ef8de07ec0ff9243391429e918b06ffe8c).

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
bdmendes: Please see [d5d001e](https://github.com/adzerk/apso/pull/1129/commits/d5d001edac59e2a7f92d56755ad88eadad774cf5).

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
jcazevedo: > Please see [1549f5e](https://github.com/adzerk/apso/pull/1129/commits/1549f5ef8de07ec0ff9243391429e918b06ffe8c).

I don't think this commit is removing the `final` keyword. Intended?

bdmendes: Meant to comment on the inner `sibling`. I'll argue for keeping the `final` for the adapter methods next.

bdmendes:
I agree that, in some cases, it might make sense to keep the possibility for implementations to override as needed for performance. However, in the cases of `/` (alias to `child`) and the `sibling` adapter I don't think it is relevant. Let me know.

modules/io/src/main/scala/com/kevel/apso/io/FileDescriptor.scala
```diff
@@ -155,7 +155,7 @@ trait FileDescriptor {
     *   the new file descriptor with the updated path
     */
   def cd(pathString: String): Self = {
-    pathString.split("/").toList.foldLeft(this.asInstanceOf[Self]) {
+    pathString.split("/").map(_.trim).toList.foldLeft(this.asInstanceOf[Self]) {
```
JD557: Since this is going to be a fold and you are touching this, I guess you can take the opportunity to avoid the allocations.


```suggestion
    pathString.split("/").iterator.map(_.trim).foldLeft(this.asInstanceOf[Self]) {
```

jcazevedo:
> I agree that, in some cases, it might make sense to keep the possibility for implementations to override as needed for performance. However, in the cases of `/` (alias to `child`) and the `sibling` adapter I don't think it is relevant. Let me know.

Fair enough.

jcazevedo (approved):
Thank you!

PR #1133, opened 2026-07-22
Use global common settings in build.sbt

bdmendes:
As of sbt2, I don't think we gain much in repeating settings across modules as root settings are now propagated to all projects. I opted to pull the settings up so that the build file is settings then modules.

Related to https://github.com/adzerk/apso/pull/1093.

No.

### How has this been tested?

Ran and compiled locally. Hopefully, this does not break the release process. 😄

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,5 @@
-import Dependencies._
-import ReleaseTransformations._
+import Dependencies.*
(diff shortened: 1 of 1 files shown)


rafaavc (approved):
Looks great, thanks!

PR #1143, opened 2026-07-28
Format build files

jcazevedo:
#1133 didn't include the changes from #1120 and now every build is failing.

Follow-up to #1133.

No.

### How has this been tested?

I'm relying on a green CI build.


bdmendes (approved):
Thanks!

PR #1149, opened 2026-08-03
Handle throttling in `FileDescriptor` implementations

rafaavc:
This PR:
- Adds `Retry#exponentialBackOff`.
- Updates `S3FileDescriptor` and `GCSFileDescriptor` to use exponential backoff when retrying.
- Fixes `ex.isRetryable` queries in `GCSBucket`.
- `S3FileDescriptor` now retries throttling exceptions by default (disable with `aws.s3.retry-on-slow-down = false`).

NA

Yes - documented the typesafe config keys of the `S3FileDescriptor`.

### How has this been tested?
Added `RetrySpec` and `S3BucketSpec`.


modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
```diff
@@ -413,10 +421,30 @@ class S3Bucket(
...
+        onMaxRetriesReached = ex => logger.error("Max retries reached. Aborting S3 operation", ex)
+      )(f)
+      .toOption
+
   private def log(isError: Boolean, message: String, cause: Throwable): Unit =
     if (isError) logger.error(message, cause) else logger.warn(message, cause)
 
   private[this] def handler: PartialFunction[Throwable, Boolean] = {
+    // Matched ahead of the shape-specific cases below, since a slow down is reported both as a service error and as a
+    // client-side error, depending on the client in use.
```
pmscosta: If it's not too much of an hassle, can we leave it a small note regarding the behavior with the CRT client? That we observed with this being not a retryable exception.

modules/core/src/main/scala/com/kevel/apso/Retry.scala
```diff
@@ -73,4 +74,86 @@ object Retry {
...
+  private[apso] def exponentialBackOffDelay(
+      attempt: Int,
+      base: FiniteDuration,
+      max: Option[FiniteDuration] = None,
+      factor: Double = 2.0,
+      jitter: FiniteDuration = 1.second
+  ): FiniteDuration = {
+    val exponential = (base.toMillis * Math.pow(factor, attempt.toDouble)).toLong
+    val capped = max.fold(exponential)(m => Math.min(exponential, m.toMillis))
+    (capped + (Random.nextDouble() * jitter.toMillis).toLong).millis
```
pmscosta: I wonder about this jitter approach. Reading through https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/, I think we could do better.

The default is 1.0 second currently and it's add after the capped value, so, to deal with contention from multiple callers, it will kind-of evenly spread things out at a maximum of like 1sec at the maximum, so for hundreds of callers, they will fall on the same second. 

There's a few different ways we can take it. If we commit to a "full jitter" implementation we can simplify this a lot and remove the `jitter` argument and just do something like: `ThreadLocalRandom.current().nextLong(capMillis + 1).millis`

However, if we want to be able to toggle the _jitterness_, maybe something like:
```scala
    val jitterFraction = Math.min(1.0, Math.max(0.0, jitter)) // being extra careful, lol...

    (
      (1.0 - jitterFraction) * capped +
        ThreadLocalRandom.current().nextDouble() * jitterFraction * capped
    ).toLong.millis
```

This seems like a better alternative, but I didn't test or benchmarks it. Thoughts?

pmscosta: Reading the javadocs of `ThreadLocalRandom` leads me to think that we want to prefer it here `ThreadLocalRandom.current()`.

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
```diff
@@ -450,25 +479,32 @@ class S3Bucket(
...
     case b: S3Bucket => b.bucketName == bucketName
     case _           => false
   }
 }
+
+object S3Bucket extends LazyLogging {
+  private[aws] val BaseBackOff = 3.seconds
+  private[aws] val MaxBackOff = 30.seconds
+  private[aws] val BackOffJitter = 1.second
+
```
pmscosta: We could consider making these values also configurable, via HOCON, in the scenario we detect we might be "stuck" a long time retrying things.

I wouldn't block merging this PR because of that, as I think these new settings, with the default of two retries, don't change the status quo.

modules/core/src/main/scala/com/kevel/apso/Retry.scala
```diff
@@ -73,4 +74,86 @@ object Retry {
...
+        case res @ Success(_)      => res
+        case failure @ Failure(ex) =>
+          if (!retryWhen(ex)) failure
+          else if (attempt >= maxRetries) {
+            onMaxRetriesReached(ex)
+            failure
+          } else {
+            val delay = exponentialBackOffDelay(attempt, base, max, factor, jitter)
+            onRetry(ex, delay, maxRetries - attempt)
+            Thread.sleep(delay.toMillis)
```
pmscosta: I came across this blog https://blog.damavis.com/en/blocking-calls-and-asynchronous-programming-with-scala/ and thought to share.

I don't think it matter that much for the current state of things, as apso is providing a "sync" API so this wouldn't have an effect on the apps using it.

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
```diff
@@ -413,10 +421,30 @@ class S3Bucket(
...
+    Retry
+      .exponentialBackOff(
+        maxRetries = maxRetries,
+        base = S3Bucket.BaseBackOff,
+        max = Some(S3Bucket.MaxBackOff),
+        jitter = S3Bucket.BackOffJitter,
+        retryWhen = !handler(_),
+        onRetry = (ex, delay, remaining) =>
+          logger.warn(s"Error during S3 operation. Retrying in ${delay.toMillis}ms ($remaining more times)", ex),
+        onMaxRetriesReached = ex => logger.error("Max retries reached. Aborting S3 operation", ex)
```
pmscosta: Doesn't this mean that we are now logging the full exception three times? On handler, onRetry, and onMaxRetriesReached. This might be costly. 

We could just log it on the handler.

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
```diff
@@ -413,10 +421,30 @@ class S3Bucket(
...
   }
 
+  private[aws] def retry[T](f: => T, maxRetries: Int = 2): Option[T] =
+    Retry
+      .exponentialBackOff(
+        maxRetries = maxRetries,
+        base = S3Bucket.BaseBackOff,
+        max = Some(S3Bucket.MaxBackOff),
+        jitter = S3Bucket.BackOffJitter,
+        retryWhen = !handler(_),
```
pmscosta: `handler` is a partial function (it already was before this PR), but it's being called normally. It might make sense to add a safeguard here, something like: `ex => !handler.applyOrElse(ex, (_: Throwable) => false)`, to make a "bug" fail-fast.

pmscosta (changes requested):

modules/core/src/main/scala/com/kevel/apso/Retry.scala
rafaavc: Great suggestion! I hadn't thought much about the jitter since this was mostly copied over from another internal project. See https://github.com/adzerk/apso/pull/1149/commits/8397dc19613f14a86ac583731eaacb1928c61f84

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
rafaavc: See https://github.com/adzerk/apso/pull/1149/commits/c91a4358e6190c8647645d3b4ee55dd9fcb58c18

modules/core/src/main/scala/com/kevel/apso/Retry.scala
rafaavc: Agreed, see https://github.com/adzerk/apso/pull/1149/commits/8397dc19613f14a86ac583731eaacb1928c61f84

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
rafaavc: See https://github.com/adzerk/apso/pull/1149/commits/3c2395dd75dfd36e447ee95088f34fcefe9cfb71

modules/core/src/main/scala/com/kevel/apso/Retry.scala
rafaavc: Yes, it's not ideal. Since this was already the behavior of the Retry module and the API is sync, I think it should be acceptable. I added the `blocking` hint in https://github.com/adzerk/apso/pull/1149/commits/3c2395dd75dfd36e447ee95088f34fcefe9cfb71

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
rafaavc: See https://github.com/adzerk/apso/pull/1149/commits/3c2395dd75dfd36e447ee95088f34fcefe9cfb71

modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
rafaavc: See https://github.com/adzerk/apso/pull/1149/commits/3c2395dd75dfd36e447ee95088f34fcefe9cfb71

pmscosta (dismissed):

modules/core/src/main/scala/com/kevel/apso/Retry.scala
rafaavc: Actually, let me remove the blocking hint so we maintain current behavior.

pmscosta (approved):

PR #1160, opened 2026-08-19
Add invalidation methods to cached functions

bdmendes:
Some users might want to invalidate entries manually, instead of waiting for the cached time window to end. In the case of async functions, the method is synchronous, but the underlying implementation seems to just remove the `Future` entry from the Caffeine map.

The current build compiles the README under `apso-docs`, which is incorrect. (Might have broken as part of the sbt2 upgrade or lift to the `modules` subdir.) I'm sneaking a fix for that here so that I can update the README with this change.

No.

Yes.

### How has this been tested?

Two new examples and looked at the generated sources.

### DIFF
--- a/modules/caching/src/main/boilerplate/com/kevel/apso/caching/MemoizeFn.scala.template
+++ b/modules/caching/src/main/boilerplate/com/kevel/apso/caching/MemoizeFn.scala.template
@@ -10,26 +10,28 @@ import MemoizeFn._
   * Provides a memoization mechanism for synchronous functions with 0 arguments using Caffeine.
   *
   * Provides a corresponding `apply()` method that follows the signature of the provided `fn` function, but is
-  * memoized using Caffeine.
+  * memoized using Caffeine. The underlying cached result can be evicted via `invalidate`.
   */
 sealed trait SyncMemoizeFn0[V] extends (() => V) {
   protected def cacheConfig: config.Cache
   protected def fn: Unit => V
   private lazy val cache: LoadingCache[Unit, V] = cacheConfig.build.build(fn)
   override def apply(): V = cache.get(())
+  def invalidate(): Unit = cache.invalidate(())
 }
 
 /**
   * Provides a memoization mechanism for asynchronous functions with 0 arguments using Caffeine.
   *
   * Provides a corresponding `apply()` method that follows the signature of the provided `fn` function, but is
-  * memoized using Caffeine.
+  * memoized using Caffeine. The underlying cached result can be evicted via `invalidate`.
   */
 sealed trait AsyncMemoizeFn0[V] extends (() => Future[V]) {
   protected def cacheConfig: config.Cache
   protected def fn: Unit => Future[V]
   private lazy val cache: AsyncLoadingCache[Unit, V] = cacheConfig.build.buildAsyncFuture(fn)
   override def apply(): Future[V] = cache.get(())
+  def invalidate(): Unit = cache.synchronous().invalidate(())
 }
 
 [#
@@ -37,13 +39,14 @@ sealed trait AsyncMemoizeFn0[V] extends (() => Future[V]) {
   * Provides a memoization mechanism for synchronous functions with 1 arguments using Caffeine.
   *
   * Provides a corresponding `apply()` method that follows the signature of the provided `fn` function, but is
-  * memoized using Caffeine.
+  * memoized using Caffeine. Underlying cached entries can be evicted via `invalidate`.
   */
 sealed trait SyncMemoizeFn1[[#I0#], V] extends Function1[[#I0#], V] {
   def cacheConfig: config.Cache
   def fn: ([#I0#]) => V
   private lazy val cache: LoadingCache[([#I0#]), V] = cacheConfig.build.build(fn.tupled)
   override def apply([#i0: I0#]): V = cache.get(([#i0#]))
+  def invalidate([#i0: I0#]): Unit = cache.invalidate(([#i0#]))
 }#
 ]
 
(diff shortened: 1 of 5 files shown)


jcazevedo:
> The current build compiles the README under `apso-docs`, which is incorrect. (Might have broken as part of the sbt2 upgrade or lift to the `modules` subdir.) I'm sneaking a fix for that here so that I can update the README with this change.

Thanks for this! During the migration to sbt 2 we changed the `mdocOut` setting from `(This Build / baseDirectory).value` to `baseDirectory.value` due to the removal of `ThisBuild`. However, `baseDirectory.value` inside `docs.settings` resolves to `apso-docs/`, which is not what we want.

modules/caching/src/test/scala/com/kevel/apso/caching/CachedFunctionsExtrasSpec.scala
```diff
@@ -60,6 +60,28 @@ class CachedFunctionsExtrasSpec(implicit ee: ExecutionEnv) extends Specification
...
       }
 
+      "evicting a key after explicit invalidation" in {
+        val counter = new AtomicInteger(0)
+        val f = () => counter.getAndIncrement()
+        val cachedF = f.cachedSync(config.Cache(Some(1.day), None))
+        cachedF() must beEqualTo(0)
+        cachedF() must beEqualTo(0)
+        cachedF.invalidate()
+        cachedF() must beEqualTo(1)
```
jcazevedo: What do you think of changing this test a little to make it evident that `invalidate()` works for single keys? Something like the following?

```suggestion
        val counter = new AtomicInteger(0)
        val f: Int => Int = x => x + counter.getAndIncrement()
        val cachedF = f.cachedSync(config.Cache(Some(1.day), None))
        cachedF(1) must beEqualTo(1)
        cachedF(1) must beEqualTo(1)
        cachedF(2) must beEqualTo(3)
        cachedF.invalidate(1)
        cachedF(1) must beEqualTo(3)
        cachedF(2) must beEqualTo(3)
        cachedF.invalidate(2)
        cachedF(1) must beEqualTo(3)
        cachedF(2) must beEqualTo(5)
```

modules/caching/src/main/boilerplate/com/kevel/apso/caching/MemoizeFn.scala.template
```diff
@@ -10,40 +10,44 @@ import MemoizeFn._
...
   * Provides a corresponding `apply()` method that follows the signature of the provided `fn` function, but is
-  * memoized using Caffeine.
+  * memoized using Caffeine. The underlying cached result can be evicted via `invalidate`.
   */
 sealed trait SyncMemoizeFn0[V] extends (() => V) {
   protected def cacheConfig: config.Cache
   protected def fn: Unit => V
   private lazy val cache: LoadingCache[Unit, V] = cacheConfig.build.build(fn)
   override def apply(): V = cache.get(())
+  def invalidate(): Unit = cache.invalidate(())
```
jcazevedo, line 20: For arity 0, both `invalidate()` and `invalidateAll()` have the same semantics. However, I think we should still provide an `invalidateAll()` method for `SyncMemoizeFn0` and `AsyncMemoizeFn0` to make the API uniform (and since it doesn't add extra complexity).

modules/caching/src/test/scala/com/kevel/apso/caching/CachedFunctionsExtrasSpec.scala
```diff
@@ -60,6 +60,28 @@ class CachedFunctionsExtrasSpec(implicit ee: ExecutionEnv) extends Specification
...
+        val cachedF = f.cachedSync(config.Cache(None, None))
+        cachedF(1) must beEqualTo(1)
+        cachedF(2) must beEqualTo(3)
+        cachedF(2) must beEqualTo(3)
+        cachedF.invalidateAll()
+        cachedF(1) must beEqualTo(3)
+        cachedF(2) must beEqualTo(5)
+      }
+
       "that resorts to the `hashCode` method of the key" in {
```
jcazevedo, line 105: I think it's helpful to also have a test for the async variant that shows what happens when a key is invalidated while the corresponding `Future` is still in-flight.

Something like the following maybe?

```scala
"evicting an in-flight computation without cancelling it" in {
  val counter = new AtomicInteger(0)
  val firstComputation = Promise[Int]()
  val secondComputation = Promise[Int]()
  val f: Int => Future[Int] = _ =>
    counter.getAndIncrement() match {
      case 0 => firstComputation.future
      case 1 => secondComputation.future
      case n => Future.failed(new IllegalStateException(s"Unexpected computation $n"))
    }
  val cachedF = f.cachedAsync(config.Cache(None, None))

  val firstCall = cachedF(1)
  val concurrentCall = cachedF(1)
  counter.get() must beEqualTo(1)

  cachedF.invalidate(1)
  val replacementCall = cachedF(1)
  counter.get() must beEqualTo(2)

  firstComputation.success(1)
  firstCall must beEqualTo(1).await
  concurrentCall must beEqualTo(1).await
  replacementCall.isCompleted must beFalse
  cachedF(1).isCompleted must beFalse
  counter.get() must beEqualTo(2)

  secondComputation.success(2)
  replacementCall must beEqualTo(2).await
  cachedF(1) must beEqualTo(2).await
  counter.get() must beEqualTo(2)
}
```

jcazevedo (changes requested):
Thanks for this! Overall this looks good! I just commented on some potential improvements to the tests.

modules/caching/src/main/boilerplate/com/kevel/apso/caching/MemoizeFn.scala.template
bdmendes: I thought about that, but ended up not adding it since I tend to like having a single way of doing things when possible (and here it's not possible to generalize any operation over cached functions of multiple arities so that we could call `invalidateAll` blindly... looking at https://docs.scala-lang.org/scala3/reference/experimental/tupled-function.html...). But I see your point, and to be honest I don't feel that strongly about this matter. Went ahead and added it in [b7648b3](https://github.com/adzerk/apso/pull/1160/commits/b7648b3ed9e1c144eb21ee51f17c776c24a7582e).

modules/caching/src/test/scala/com/kevel/apso/caching/CachedFunctionsExtrasSpec.scala
bdmendes: Thanks! Added that test and some more in [38c9751](https://github.com/adzerk/apso/pull/1160/commits/38c975172197090420e8d4bd6c8348846794e465).

jcazevedo (approved):
Thank you! Looks great!

PR #1167, opened 2026-08-24
Work around duplicate sbt-boilerplate generation under sbt2 and bump version

bdmendes:
This release (first after the upgrade to sbt2) had some hiccups.

```
[error] (caching / Compile / packageSrc) java.util.zip.ZipException: duplicate entry: com/kevel/apso/caching/CachedFunctionsExtras.scala
```

Also, for future reference, for `sbt-release` to work properly in sbt2, `version` must still be scoped under `ThisBuild`, which we removed manually in https://github.com/adzerk/apso/commit/a20021712b47bd6ef0e54fe3b54f2f29d4c4c8ed.

### Does this change relate to existing issues or pull requests?

No. This fixes the release process.

### Does this change require an update to the documentation?

No.

### How has this been tested?

Releasing 0.27.5.

PR #1168, opened 2026-08-24
Ensure publish process works in CI

bdmendes:
Title. This should give us more confidence when updating the build.

Suggested in https://github.com/adzerk/apso/pull/1167#pullrequestreview-5007780101

No.

### How has this been tested?

N/A


jcazevedo (approved):
Thank you!

PR #1182, opened 2026-09-09
Prefer CopyRequest for `move` API

pmscosta:
We detected livelock issue with using the AWS S3 CRT client in our current move implementation.
It seemed to be caused by a CRT upload fed by a blocking InputStream that is itself created by another CRT download with the same client.
If they both land on the same _event loop_ and the object is large (from our testing, above 16MB), the upload loops infinitely and stops the download from fetching data.

Follow-up on #1128 

No.

### How has this been tested?

Local ad-hoc testing. Verified this caused no deadlocks when moving files agains a MinIO backend (since the problem is a logic one with aws).

This also seems to provide more throughput. I didn't measure against a true S3 bucket, but against MinIO backend.

this branch:
```
scenario: server-copy; size: 512 MiB; runs: 5
copy 1: 0.498s, 1028.0 MiB/s
copy 2: 0.442s, 1159.3 MiB/s
copy 3: 0.312s, 1639.8 MiB/s
copy 4: 0.364s, 1408.4 MiB/s
copy 5: 0.313s, 1636.7 MiB/s
mean: 1374.4 MiB/s
```

main:
```
scenario: streamed-move; size: 512 MiB; runs: 5
copy 1: 2.980s, 171.8 MiB/s
copy 2: 2.899s, 176.6 MiB/s
copy 3: 2.799s, 182.9 MiB/s
copy 4: 2.738s, 187.0 MiB/s
copy 5: 2.762s, 185.3 MiB/s
mean: 180.7 MiB/s
```


modules/aws/src/main/scala/com/kevel/apso/aws/S3Bucket.scala
```diff
@@ -379,6 +379,42 @@ class S3Bucket(
...
+            CopyObjectRequest
+              .builder()
+              .sourceBucket(bucketName)
+              .sourceKey(sanitizeKey(sourceKey))
+              .destinationBucket(bucketName)
+              .destinationKey(sanitizeKey(destinationKey))
+              .build()
+          )
+          .build()
+      ).completionFuture().join()
```
bdmendes, line 414: It is a bummer that the current API is synchronous; we should definitely invest some time in the future to come up with a thin NIO wrapper or something like that.

bdmendes (approved):
Thanks!

